PR #3, opened 2026-07-17
Adapt Cluster Autoscaler to be the only component in the new repo

towca:
#### What this PR does / why we need it:

This PR is a part of the first mechanical steps of migrating core Cluster Autoscaler logic from [its existing repository](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) to a library in a new repository (this one).

Before submitting this PR, I force-pushed the `main` branch of this repository to the following starting point:
* Started with checking out the master branch from [the existing kubernetes/autoscaler repo](https://github.com/kubernetes/autoscaler). This repository contains multiple components, including Cluster Autoscaler.
* Filtered the history to only contain files and commits related to Cluster Autoscaler by running:
  ```
  git filter-repo --path .github/ \
                  --path cluster-autoscaler/ \
                  --path hack/ \
                  --path .gitignore \
                  --path .pre-commit-config.yaml \
                  --force
  ``` 
* Created an empty commit signifying the end of the history from the old repo.
* Cherry-picked the commits that were already present in this repo, with some manual changes incorporating the files from the old repo.

This PR contains a series of follow-up changes aimed at making Cluster Autoscaler the only component in the new repo. These changes didn't need to be force-pushed and can go through a normal review process. The details about individual changes can be found in commit messages.

#### Special notes for your reviewer:

The PR is organized into meaningful commits to be reviewed sequentially. Some of the commits have huge diffs because they modify most files in the repo - but all such commits make identical modifications to 99% of the files, to make the review feasible. Reviewing these huge commits through GitHub might not work well - I'd recommend checking the PR out using `gh pr checkout`, and reviewing it directly via git.

Changes to the old repo haven't been frozen yet, so the force-push part described above will have to be redone before submitting this PR. Specifically, this PR should be submitted in the following way:

1. Freeze changes to [the old repo](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler).
1. Redo force-pushing the `main` from this repo to contain the history of the `master` from the old repo (described above).
1. Create and merge a PR in the the old repo, which removes the core logic files that are intended to stay in this repo.
1. Unfreeze changes to the old repo.
1. Rebase this PR on top of the newly-force-pushed `main`, handle conflicts in the files that were changed in the old repo since this PR was first created. Submit the PR.

Leaving a hold so that the PR doesn't get merged accidentally.
/hold

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

[...]

PR #4, opened 2026-07-20
Adapt Cluster Autoscaler to be the only component in the new repo

Choraden:
This is a clone of #3 with some extra fixes on top.

This PR is a part of the first mechanical steps of migrating core Cluster Autoscaler logic from [its existing repository](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) to a library in a new repository (this one).

Before submitting this PR, I force-pushed the `main` branch of this repository to the following starting point:
* Started with checking out the master branch from [the existing kubernetes/autoscaler repo](https://github.com/kubernetes/autoscaler). This repository contains multiple components, including Cluster Autoscaler.
* Filtered the history to only contain files and commits related to Cluster Autoscaler by running:
  ```
  git filter-repo --path .github/ \
                  --path cluster-autoscaler/ \
                  --path hack/ \
                  --path .gitignore \
                  --path .pre-commit-config.yaml \
                  --force
  ``` 
* Created an empty commit signifying the end of the history from the old repo.
* Cherry-picked the commits that were already present in this repo, with some manual changes incorporating the files from the old repo.

This PR contains a series of follow-up changes aimed at making Cluster Autoscaler the only component in the new repo. These changes didn't need to be force-pushed and can go through a normal review process. The details about individual changes can be found in commit messages.

The PR is organized into meaningful commits to be reviewed sequentially. Some of the commits have huge diffs because they modify most files in the repo - but all such commits make identical modifications to 99% of the files, to make the review feasible. Reviewing these huge commits through GitHub might not work well - I'd recommend checking the PR out using `gh pr checkout`, and reviewing it directly via git.

Changes to the old repo haven't been frozen yet, so the force-push part described above will have to be redone before submitting this PR. Specifically, this PR should be submitted in the following way:

1. Freeze changes to [the old repo](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler).
1. Redo force-pushing the `main` from this repo to contain the history of the `master` from the old repo (described above).
1. Create and merge a PR in the the old repo, which removes the core logic files that are intended to stay in this repo.
1. Unfreeze changes to the old repo.
1. Rebase this PR on top of the newly-force-pushed `main`, handle conflicts in the files that were changed in the old repo since this PR was first created. Submit the PR.

Leaving a hold so that the PR doesn't get merged accidentally.
/hold

```release-note
NONE
```

* Issue tracking the whole migration of core Cluster Autoscaler logic into a library in a new repo: https://github.com/kubernetes/autoscaler/issues/9264
* Issue tracking the first mechanical steps of the migration - splitting files/logic between two repos: https://github.com/kubernetes/autoscaler/issues/9832


Choraden:
The benchmark job fails because the base (current main) has different files structure than introduced in the pr (`cluster-autoscaler/` -> `pkg/`). I verified the job manually.

ameukam:
Out of curiosity, why not put `cluster-autoscaler/*` at the root of the repo ? is it due to the git plugin ?

Choraden:
>Out of curiosity, why not put cluster-autoscaler/* at the root of the repo ? is it due to the git plugin ?

@ameukam 

We decided to move all the files from `cluster-autoscaler/` to `pkg/` manually with a commit, instead of using filter-plugin to preserve building / working tests in case of debugging of an old version.

Also we found your suggestion to use `pkg/` a good idea.

ameukam:
FWIW,
/lgtm

elmiko:
i checked this out locally and poked around, it seems good to me but i didn't test the built binaries.

BigDarkClown:
Looks reasonable from my side.

jackfrancis:
@Choraden do we want to remove the Benchmark GitHub Action job and file an issue to add it back at some point in the future?

Choraden:
@jackfrancis I believe that's what we need to do. I mean according to the comment above:

> The benchmark job fails because the base (current main) has different files structure than introduced in the pr (cluster-autoscaler/ -> pkg/).

once we merge this PR, it will become operational again. The problem is that we probably won't be able to merge it with failed job, right? 😄 I guess, I'll just drop the benchmark and have it added later.

jackfrancis:
I'm happy to merge and then fix later. There's no way to avoid this process being messy at this stage. :)

Permission to merge?

Choraden:
@jackfrancis Please don't merge. 

We need to force push main branch first to match the HEAD of k/a and simultaneously merge kubernetes/autoscaler#10002. There is a high chance me and @towca will do it tomorrow/Friday.

.github/dependabot.yml
BigDarkClown: Missed dir rename?

Choraden: Good catch. Done.

PR #18, opened 2026-08-03
Benchmarking scaledown strategies

viktoriaas:
#### What type of PR is this?

/kind feature

This project aims to develop a system for benchmarking various scaledown strategies and evaluating how chosen strategy affects final efficiency. Scaledown strategies may be driven by different algorithms and prioritization logic, therefore it is essential to first develop metrics that will assess the scaledown outcome objectively.

The initial work involves:

- Definition of `scaleDownMetric` interface and `metricsTracker` to orchestrate the computation, storing, and reporting of multiple `scaleDownMetrics`.
- Implementation of  `resourceUtilizationMetric`.
- Implementation of `fakeActuator` which performs node removal from clustersnapshot to simulate scaledown actuation without external cloud provider communication.
- Definition of `scaleDownScenario` which packs initial cluster topology, `metricsTracker` and optional `AutoscalingOptions` into one struct for individual scaledown run execution.
- Implementation of benchmark runner in the form of a simplified scaledown loop which executes until no more nodes to delete are detected.

A small change was introduced to `simulator/clustersnaphot, func InitializeClusterSnapshotOrDie` - parameter `t *testing.T` changed to `t testing.TB` to enable reusing the function from the benchmark.

#### Run 
Benchmark can be run from cmdl e.g. 
```
cd pkg/core/bench/scaledown/efficiency
go test -v -count=5 -benchtime=1x -bench=.  .
```
which will run the scaledown loop until no new scaledown candidates are found 5 times on fresh initial cluster state and report raw metrics for these 5 executions. If the output is redirected to results file, it can be fed to benchstat.

An explanation to why `benchtime=1x` is used is offered in the [comment to the PR in previous CA repository](https://github.com/kubernetes/autoscaler/pull/9986#discussion_r3676034501). 

This work has been previously tracked in kubernetes/autoscaler repository.
```docs
[PR 9986 in kubernetes/autoscaler]: https://github.com/kubernetes/autoscaler/pull/9986
```

### DIFF
--- a/pkg/core/bench/scaledown/efficiency/benchmark_scaledown_test.go
+++ b/pkg/core/bench/scaledown/efficiency/benchmark_scaledown_test.go
@@ -1,0 +1,209 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package efficiency
+
+import (
+	"testing"
+	"time"
+
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/klog/v2"
+	testprovider "sigs.k8s.io/cluster-autoscaler/pkg/cloudprovider/test"
+	"sigs.k8s.io/cluster-autoscaler/pkg/config"
+	cacontext "sigs.k8s.io/cluster-autoscaler/pkg/context"
+	"sigs.k8s.io/cluster-autoscaler/pkg/core/bench/scaledown/efficiency/metrics"
+	"sigs.k8s.io/cluster-autoscaler/pkg/core/scaledown/budgets"
+	"sigs.k8s.io/cluster-autoscaler/pkg/core/scaledown/planner"
+	. "sigs.k8s.io/cluster-autoscaler/pkg/core/test"
+	"sigs.k8s.io/cluster-autoscaler/pkg/processors"
+	processorstest "sigs.k8s.io/cluster-autoscaler/pkg/processors/test"
+	"sigs.k8s.io/cluster-autoscaler/pkg/resourcequotas"
+	"sigs.k8s.io/cluster-autoscaler/pkg/simulator/clustersnapshot"
+	"sigs.k8s.io/cluster-autoscaler/pkg/simulator/options"
+	kubeutil "sigs.k8s.io/cluster-autoscaler/pkg/utils/kubernetes"
+)
+
+func init() {
+	klog.InitFlags(nil)
+}
+
+// scaleDownScenario packs initialClusterState, metricsTracker, and optional config.AutoscalingOptions into one struct for individual scaledown run execution.
+type scaleDownScenario struct {
+	Name                 string
+	InitialClusterState  *initialClusterState
+	MetricsTracker       *metrics.Tracker
+	AutoscalingOptsSetup func(opts *config.AutoscalingOptions)
+}
+
+type scaleDownDependencies struct {
+	AutoscalingCtx *cacontext.AutoscalingContext
+	Processors     *processors.AutoscalingProcessors
+	Planner        *planner.Planner
+	Actuator       *fakeActuator
+	MetricsTracker *metrics.Tracker
+}
+
+func defaultAutoscalingOptions() config.AutoscalingOptions {
+	return config.AutoscalingOptions{
+		ScaleDownEnabled:           true,
+		MaxScaleDownParallelism:    1,
+		MaxDrainParallelism:        1,
+		ScaleDownSimulationTimeout: 10 * time.Second,
+		NodeGroupDefaults: config.NodeGroupAutoscalingOptions{
+			ScaleDownUtilizationThreshold: 1.0,
+			ScaleDownUnneededTime:         0,
+		},
+	}
+}
+
+func buildScaleDownDependencies(b *testing.B, cs *initialClusterState, autoscalingOpts config.AutoscalingOptions) scaleDownDependencies {
+	b.Helper()
+	var allNodes []*apiv1.Node
+	var allPods []*apiv1.Pod
+
+	provider := testprovider.NewTestCloudProviderBuilder().Build()
+
+	for _, ng := range cs.NodeGroups {
+		allNodes = append(allNodes, ng.Nodes...)
+		allPods = append(allPods, ng.Pods...)
+		provider.AddNodeGroupWithCustomOptions(ng.Name, ng.MinSize, ng.MaxSize, len(ng.Nodes), ng.NodeGroupOptions)
+		for _, node := range ng.Nodes {
+			provider.AddNode(ng.Name, node)
+		}
+	}
+
+	procs, templateRegistry := processorstest.NewTestProcessors(autoscalingOpts)
+	registry := kubeutil.NewListerRegistry(nil, nil, nil, nil, nil, nil, nil, nil, nil)
+
+	autoscalingCtx, err := NewScaleTestAutoscalingContext(autoscalingOpts, &fake.Clientset{}, registry, provider, nil, nil, templateRegistry)
+	if err != nil {
+		b.Fatalf("failed to instantiate AutoscalingContext: %v", err)
+	}
+	clustersnapshot.InitializeClusterSnapshotOrDie(b, autoscalingCtx.ClusterSnapshot, allNodes, allPods)
+
+	a := NewFakeActuator(autoscalingCtx, &fakeActuationStatus{})
+	factory := resourcequotas.NewTrackerFactory(resourcequotas.TrackerOptions{CustomResourcesProcessor: procs.CustomResourcesProcessor, QuotaProvider: resourcequotas.NewCloudMinProvider(provider)})
+	p := planner.New(&autoscalingCtx, procs, options.NodeDeleteOptions{}, nil, factory)
+
+	return scaleDownDependencies{
+		AutoscalingCtx: &autoscalingCtx,
+		Processors:     procs,
+		Planner:        p,
+		Actuator:       a,
+	}
+}
+
+// Crucial to pass -benchtime=1x to let the Loop know to run scaledown strategy exactly once.
+func (scenario scaleDownScenario) run(b *testing.B) {
+	opts := defaultAutoscalingOptions()
+	if scenario.AutoscalingOptsSetup != nil {
+		scenario.AutoscalingOptsSetup(&opts)
+	}
+	mt := scenario.MetricsTracker
+	sdDeps := buildScaleDownDependencies(b, scenario.InitialClusterState, opts)
+
+	nodeInfos, err := sdDeps.AutoscalingCtx.ClusterSnapshot.ListNodeInfos()
+	if err != nil {
+		b.Fatalf("failed to list node infos from snapshot: %s", err.Error())
+	}
+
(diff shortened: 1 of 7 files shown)


pkg/core/bench/scaledown/efficiency/benchmark_scaledown_test.go
```diff
@@ -0,0 +1,211 @@
...
+		ScaleDownSimulationTimeout: 10 * time.Second,
+		NodeGroupDefaults: config.NodeGroupAutoscalingOptions{
+			ScaleDownUtilizationThreshold: 1.0,
+			ScaleDownUnneededTime:         0,
+		},
+	}
+}
+
+func buildScaleDownDependencies(b *testing.B, cs *initialClusterState, autoscalingOpts config.AutoscalingOptions) scaleDownDependencies {
+	var allNodes []*apiv1.Node
```
norbertcyran, line 76: nit: call `b.Helper()` here

pkg/core/bench/scaledown/efficiency/benchmark_scaledown_test.go
```diff
@@ -0,0 +1,211 @@
...
+		if sdDeps.Processors == nil || sdDeps.Processors.ScaleDownNodeProcessor == nil {
+			scaleDownCandidates = currentNodes
+			podDestinations = currentNodes
+		} else {
+			scaleDownCandidates, err = sdDeps.Processors.ScaleDownNodeProcessor.GetScaleDownCandidates(
+				sdDeps.AutoscalingCtx, currentNodes)
+			if err != nil {
+				b.Fatalf("error getting scale-down candidates: %s", err.Error())
+			}
+		}
```
norbertcyran, line 154: If I'm not mistaken, we don't populate `podDestinations` in this branch

pkg/core/bench/scaledown/efficiency/benchmark_scaledown_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+		needDrain = []*apiv1.Node{}
+		for _, group := range drainGrouped {
+			needDrain = append(needDrain, group.Nodes...)
+		}
+
+		// No more nodes to delete, final state reached.
+		if len(empty) == 0 && len(needDrain) == 0 {
+			b.ReportMetric(float64(scaleDownLoop), "loops")
```
norbertcyran, line 179: nit: I think `scaleDownLoop` should start from 0 because we break the loop before the scale down actuation kicks in

pkg/core/bench/scaledown/efficiency/fake_actuator.go
```diff
@@ -0,0 +1,144 @@
...
+func (f *fakeActuator) startDeletion(empty, needDrain []*apiv1.Node) (status.ScaleDownResult, []*status.ScaleDownNode, caerrors.AutoscalerError) {
+	if len(empty) == 0 && len(needDrain) == 0 {
+		return status.ScaleDownNoNodeDeleted, nil, nil
+	}
+
+	var scaledDownNodes []*status.ScaleDownNode
+	for _, node := range empty {
+		err := f.autoscalingCtx.ClusterSnapshot.RemoveNodeInfo(node.Name)
+		if err != nil {
+			//f.t.Logf("cannot remove node info, err: %v", err.Error())
```
norbertcyran: to be removed, I guess

pkg/core/bench/scaledown/efficiency/fake_actuator.go
```diff
@@ -0,0 +1,144 @@
...
+		if len(podsToReschedule) > 0 {
+			schedulingResult, err := schedSimulator.TrySchedulePods(context.Background(), f.autoscalingCtx.ClusterSnapshot, podsToReschedule, false,
+				clustersnapshot.SchedulingOptions{
+					IsNodeAcceptable: scheduling.ScheduleAnywhere,
+				})
+			if err != nil {
+				return status.ScaleDownError, nil, caerrors.ToAutoscalerError(caerrors.TransientError, err)
+			}
+
+			scheduledPods := make(map[string]bool)
```
norbertcyran: `map[string]bool` is totally fine here. Just FYI, there is a set implementation in k8s apimachinery: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets

pkg/core/bench/scaledown/efficiency/fake_actuator.go
```diff
@@ -0,0 +1,144 @@
...
+					if f.actuationStatus != nil {
+						f.actuationStatus.RegisterEviction(pod)
+					}
+				}
+			}
+
+			if len(schedulingResult.Statuses) != len(podsToReschedule) {
+				klog.V(5).Infof("failed to reschedule all pods from node %s: scheduled %d out of %d pods", n.Name, len(schedulingResult.Statuses), len(podsToReschedule))
+				continue
+			}
```
norbertcyran: shouldn't that be treated as an error? Has it actually happened when you were running this benchmark?

norbertcyran (commented):
overall lgtm, just a few final nits. I think it's a great start!

pkg/core/bench/scaledown/efficiency/fake_actuator.go
viktoriaas: ~~As discussed offline, the logic flow regarding `needDrain` nodes in actuator is changed to:~~

```
for N in needDrain:
  Gather all Pods to be rescheduled from N and set their nodeName to empty
  Try rescheduling these Pods
  For each Pod that was not rescheduled successfully, set its node name back to N.Name
  Iff all Pods from N were successfully rescheduled
      Remove N from clusterSnaphot
      Add it to scaledDownNodes
```

~~Partially successful drain is not error per se since planner already marked this node as suitable for removal and we act on this decision. With the modified actuator logic, we are therefore not treating partially successful drain as error but we just do not remove that node from the snapshot and don't add it to scaledDownNodes.~~

**Update:** I took another look at this and realized the original flow was correct and we should treat an unsuccessful reschedule as an error. 

The `Planner` already simulates the rescheduling for each candidate node and won't include a node as a scaledown candidate if that simulation fails. Because the fake Actuator operates on the same static cluster snapshot, nothing changes in this cluster state between the planner's simulation outcome and the fake Actuator's execution. Therefore, the drain should always be 100% successful, and any partial failure is really an error.

pkg/core/bench/scaledown/efficiency/benchmark_scaledown_test.go
viktoriaas: Indeed, thanks for pointing this out!

pkg/core/bench/scaledown/efficiency/metrics/interface.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+// ComputeMetrics computes metric value across a set of configured scaleDownMetrics.
+func (mt *metricsTracker) ComputeMetrics(clusterState clusterState, t testing.TB) {
+	for _, m := range mt.metrics {
+		val, err := m.Compute(clusterState)
+		if err != nil {
+			t.Errorf("failed to compute metric %s, err %v", m.Name(), err)
+			continue
```
BigDarkClown, line 85: This handling can cause metrics to be misaligned - if a metric returns an error we will have diverging lengths of datapoints. Have you considered how this would be later reported to the benchmark tool?

TBF, a good solution might be to change the `Compute` signature to not return an error, but instead have it return some default value.

pkg/core/bench/scaledown/efficiency/metrics.go
```diff
@@ -80,3 +80,89 @@ func (mt *metricsTracker) ComputeMetrics(clusterState clusterState, t testing.TB
 		mt.metricsTimeline[m.Name()] = append(mt.metricsTimeline[m.Name()], val)
 	}
 }
+
+// resource struct holds two values:
```
BigDarkClown: nit: A better long-term pattern is defining new metrics in a new `metrics` package, using a single file per metric. This way we won't bloat `metrics.go` in time.

pkg/core/bench/scaledown/efficiency/metrics.go
```diff
@@ -80,3 +80,89 @@ func (mt *metricsTracker) ComputeMetrics(clusterState clusterState, t testing.TB
...
+func resourceAllocatableRequested(resourceName apiv1.ResourceName, nodeInfo *framework.NodeInfo, nodeName string, currentTime time.Time) (resource, error) {
+	var r resource
+	util, err := utilization.CalculateUtilizationOfResource(nodeInfo, resourceName, true, true, currentTime)
+	if err != nil {
+		return r, err
+	}
+	r.utilization = util
+	allocatable, found := nodeInfo.Node().Status.Allocatable[resourceName]
+	if !found {
+		return r, fmt.Errorf("failed to get allocatable %v from %s", resourceName, nodeName)
```
BigDarkClown: `nodeName` is already passed to the function as `nodeInfo.Node().Name`, having it be a dedicated argument to `resourceAllocatableRequested` is redundant.

pkg/core/bench/scaledown/efficiency/metrics.go
```diff
@@ -80,3 +80,89 @@ func (mt *metricsTracker) ComputeMetrics(clusterState clusterState, t testing.TB
...
+	}
+	if resourceName == apiv1.ResourceCPU {
+		r.allocatable = float64(allocatable.MilliValue())
+	} else {
+		r.allocatable = float64(allocatable.Value())
+	}
+	return r, nil
+}
+
+func computeResourceUtilization(metricName string, resourceName apiv1.ResourceName, clusterState clusterState) (float64, error) {
```
BigDarkClown: I think both this and the other aux functions could be underneath `resourceUtilizationMetric` - this way we also won't have to pass the metricName.

pkg/core/bench/scaledown/efficiency/benchmark_scaledown_test.go
```diff
@@ -0,0 +1,211 @@
...
+func init() {
+	klog.InitFlags(nil)
+}
+
+// scaleDownScenario packs initialClusterState, metricsTracker, and optional config.AutoscalingOptions into one struct for individual scaledown run execution.
+type scaleDownScenario struct {
+	Name                string
+	InitialClusterState *initialClusterState
+	MetricsTracker      *metricsTracker
+	AutoscalingOpts     func(opts *config.AutoscalingOptions)
```
BigDarkClown: nit: `AutoscalingOptsSetup`?

pkg/core/bench/scaledown/efficiency/benchmark_scaledown_test.go
```diff
@@ -0,0 +1,211 @@
...
+	var scaleDownCandidates []*apiv1.Node
+	var podDestinations []*apiv1.Node
+
+	for scaleDownLoop := 1; ; scaleDownLoop++ {
+		infos, err := sdDeps.AutoscalingCtx.ClusterSnapshot.ListNodeInfos()
+		if err != nil {
+			b.Fatalf("error retrieving nodes from cluster snapshot, err %s", err.Error())
+		}
+
+		var currentNodes []*apiv1.Node
```
BigDarkClown: nit: just `nodes`?

BigDarkClown (commented):
meta nit: A better way to update the chain after feedback is to amend it to the commits. This way we don't have a "cleanup" commits that remove the code added in previous commits.

pkg/core/bench/scaledown/efficiency/metrics.go
viktoriaas: I moved `computeResourceUtilization` under `resourceUtilizationMetric` . The other aux function (`resourceAllocatableRequested`) should stay outside the `resourceUtilizationMetric` because it is reused for computing another upcoming metric (resource fragmentation).

pkg/core/bench/scaledown/efficiency/metrics/interface.go
viktoriaas: I'm aware of this behaviour but it has not been decided yet if such event should be treated as a hard error and the `scaleDownMetric` should not be included in final benchmark at all or if there is a suitable way of computing the final value from partial `metricsTimeline`.  

The values for different metrics can have wide range so any arbitrary number cannot serve as default. In addition, even established default does not solve the question of how to deal with partial `metricsTimeline`.

For now, I have added a comment to the `Compute` function that final `metricTimeline` may diverge in length across configured `scaleDownMetric` if the metric fails to be computed.

PR #19, opened 2026-08-03
[Scaledown] Partial taint actuation

tetianakh:
#### What type of PR is this?
/kind feature

Currently, if at least one node couldn't be tainted in a batch of nodes selected for scaledown, taints are cleaned up from all nodes and none of the nodes in the batch are scaled down.

This PR has 4 commits:

1. First commit adds a feature flag `--partial-taint-actuation-enabled`. 
2. Second commit removes a 30s wait in the latency tracker unit tests and refactors latency tracker expect specific number of nodes.
3. Third commit implements the partial taint actuation. If the flag is set to true, CA will scale down all nodes in non-atomic nodegroups that were successfully tainted.  For atomic nodegroups, the behavior is unchanged. If any nodes from an atomic nodegroup weren't tainted successfully, the taints are cleaned up from all the nodes in this nodegroup. The only change is that with partial-taint-actuation-enabled set to true, cleanup is done in parallel  (maxConcurrentNodesTainting nodes at a time). Integration tests for scaledown were added. 
4. The last commit changes how dynamic delay after taint behaves. Currently it measures the time between calling taints.MarkToBeDeleted() and the time when the node appears tainted in the nodeLister (the informer). This mechanism overestimates the latency greatly because the request can spend many seconds in the client-side queue (the queue size is controlled by --kube-api-qps and --kube-api-burst flags). This is a bug fix and it is not hidden behind the feature flag.

Fixes #

```release-note
Adds `--partial-taint-actuation-enabled` flag. If it is set to true, CA will scale down all nodes in non-atomic nodegroups that were successfully tainted. 
```

```docs

```

### DIFF
--- a/pkg/core/scaledown/actuation/actuator.go
+++ b/pkg/core/scaledown/actuation/actuator.go
@@ -235,10 +235,6 @@ func (a *Actuator) taintNodesSync(ctx context.Context, nodeGroupViews []*budgets
 	nodesToTaint := make([]*apiv1.Node, 0)
 	var updateLatencyTracker *UpdateLatencyTracker
 	nodeDeleteDelayAfterTaint := a.nodeDeleteDelayAfterTaint
-	if a.autoscalingCtx.AutoscalingOptions.DynamicNodeDeleteDelayAfterTaintEnabled {
-		updateLatencyTracker = NewUpdateLatencyTracker(a.autoscalingCtx.AutoscalingKubeClients.ListerRegistry.AllNodeLister())
-		go updateLatencyTracker.Start(ctx)
-	}
 
 	for _, bucket := range nodeGroupViews {
 		for _, node := range bucket.Nodes {
@@ -252,7 +248,7 @@ func (a *Actuator) taintNodesSync(ctx context.Context, nodeGroupViews []*budgets
 		for _, node := range nodesToTaint {
 			updateLatencyTracker.StartTimeChan <- nodeTaintStartTime{node.Name, time.Now()}
 		}
-		go updateLatencyTracker.Start()
+		go updateLatencyTracker.Start(ctx)
 	}
 
 	// Taint nodes concurrently and record failures
@@ -279,7 +275,7 @@ func (a *Actuator) taintNodesSync(ctx context.Context, nodeGroupViews []*budgets
 			if _, found := failedNodes[taintedNode.UID]; found {
 				continue
 			}
-			_, _ = taints.CleanToBeDeleted(ctx,taintedNode, a.autoscalingCtx.ClientSet, a.autoscalingCtx.CordonNodeBeforeTerminate)
+			_, _ = taints.CleanToBeDeleted(ctx, taintedNode, a.autoscalingCtx.ClientSet, a.autoscalingCtx.CordonNodeBeforeTerminate)
 		}
 		// No need to record taint propagation latency, all taints are cleaned up.
 		if a.autoscalingCtx.AutoscalingOptions.DynamicNodeDeleteDelayAfterTaintEnabled {
@@ -309,7 +305,7 @@ func (a *Actuator) taintNodesSync(ctx context.Context, nodeGroupViews []*budgets
 		klog.Infof("couldn't taint %d nodes with ToBeDeleted, proceeding with partial scale down or bucket cleanup", len(failedNodes))
 
 		for _, bucket := range nodeGroupViews {
-			bucketWithSuccessfulNodes, bucketNodesToClean := a.resolveBucketFailures(bucket, failedNodes)
+			bucketWithSuccessfulNodes, bucketNodesToClean := a.resolveBucketFailures(ctx, bucket, failedNodes)
 			if bucketWithSuccessfulNodes != nil {
 				successfulNodeGroupViews = append(successfulNodeGroupViews, bucketWithSuccessfulNodes)
 			}
@@ -334,8 +330,8 @@ func (a *Actuator) taintNodesSync(ctx context.Context, nodeGroupViews []*budgets
 //   - If the node is from a regular, non-atomic nodegorup, other successfully tainted nodes from this nodegroup can be deleted.
 //
 // Returns a nodegroup view that contains nodes that can be deleted, and a list of nodes from which the taint has to be cleaned up.
-func (a *Actuator) resolveBucketFailures(bucket *budgets.NodeGroupView, failedNodes map[types.UID]struct{}) (*budgets.NodeGroupView, []*apiv1.Node) {
-	opts, err := bucket.Group.GetOptions(a.autoscalingCtx.NodeGroupDefaults)
+func (a *Actuator) resolveBucketFailures(ctx context.Context, bucket *budgets.NodeGroupView, failedNodes map[types.UID]struct{}) (*budgets.NodeGroupView, []*apiv1.Node) {
+	opts, err := bucket.Group.GetOptions(ctx, a.autoscalingCtx.NodeGroupDefaults)
 	isAtomic := false
 	if err != nil {
 		klog.Warningf("Failed to get options for node group %v: %v, assuming atomic node group to be safe", bucket.Group.Id(), err)
--- a/pkg/core/scaledown/actuation/actuator.go
+++ b/pkg/core/scaledown/actuation/actuator.go
@@ -241,13 +241,13 @@ func (a *Actuator) taintNodesSync(ctx context.Context, NodeGroupViews []*budgets
 			_, _ = taints.CleanToBeDeleted(ctx, taintedNode, a.autoscalingCtx.ClientSet, a.autoscalingCtx.CordonNodeBeforeTerminate)
 		}
 		if a.autoscalingCtx.AutoscalingOptions.DynamicNodeDeleteDelayAfterTaintEnabled {
-			close(updateLatencyTracker.AwaitOrStopChan)
+			close(updateLatencyTracker.ExpectedNodeCountChan)
 		}
 		return nodeDeleteDelayAfterTaint, errors.NewAutoscalerErrorf(errors.ApiCallError, "couldn't taint %d nodes with ToBeDeleted", len(failedTaintedNodes))
 	}
 
 	if a.autoscalingCtx.AutoscalingOptions.DynamicNodeDeleteDelayAfterTaintEnabled {
-		updateLatencyTracker.AwaitOrStopChan <- true
+		updateLatencyTracker.ExpectedNodeCountChan <- len(nodesToTaint)
 		latency, ok := <-updateLatencyTracker.ResultChan
 		if ok {
 			a.pastLatencies.RegisterElement(latency)
--- a/pkg/core/scaledown/actuation/actuator.go
+++ b/pkg/core/scaledown/actuation/actuator.go
@@ -140,7 +140,7 @@ func (a *Actuator) startDeletion(ctx context.Context, empty, drain []*apiv1.Node
 	defer func() {
 		// nodesToClean is populated only if PartialTaintActuationEnabled is true.
 		if len(nodesToClean) > 0 {
-			workqueue.ParallelizeUntil(context.Background(), maxConcurrentNodesTainting, len(nodesToClean), func(piece int) {
+			workqueue.ParallelizeUntil(ctx, maxConcurrentNodesTainting, len(nodesToClean), func(piece int) {
 				_, _ = taints.CleanToBeDeleted(ctx, nodesToClean[piece], a.autoscalingCtx.ClientSet, a.autoscalingCtx.CordonNodeBeforeTerminate)
 			})
 		}
@@ -152,18 +152,48 @@ func (a *Actuator) startDeletion(ctx context.Context, empty, drain []*apiv1.Node
 		return status.ScaleDownNoNodeDeleted, nil, nil
 	}
 
+	abortedNodeGroups := make(map[string]bool)
+
 	if len(emptyToDelete) > 0 {
 		// Taint all empty nodes synchronously
 		taintRes, err := a.taintNodesSync(ctx, emptyToDelete)
 		if len(taintRes.nodesToClean) > 0 {
 			nodesToClean = append(nodesToClean, taintRes.nodesToClean...)
 		}
+
+		// Map successfully processed groups
+		successfulGroups := make(map[string]bool)
+		for _, bucket := range taintRes.successfulNodes {
+			successfulGroups[bucket.Group.Id()] = true
+		}
+		// If an original bucket didn't make it to successfulNodes, it was aborted.
+		// If the group is atomic, we must abort its siblings in the drain phase.
+		for _, bucket := range emptyToDelete {
+			if !successfulGroups[bucket.Group.Id()] {
+				opts, err := bucket.Group.GetOptions(ctx, a.autoscalingCtx.NodeGroupDefaults)
+				if err != nil || (opts != nil && opts.ZeroOrMaxNodeScaling) {
+					abortedNodeGroups[bucket.Group.Id()] = true
+				}
+			}
+		}
+
 		if err != nil {
 			return status.ScaleDownError, scaledDownNodes, err
 		}
 
 		emptyScaledDown := a.deleteAsyncEmpty(ctx, taintRes.successfulNodes, taintRes.delayAfterTaint, force)
 		scaledDownNodes = append(scaledDownNodes, emptyScaledDown...)
+	}
+
+	if len(drainToDelete) > 0 {
+		var filteredDrainToDelete []*budgets.NodeGroupView
+		for _, bucket := range drainToDelete {
+			if abortedNodeGroups[bucket.Group.Id()] {
+				continue
+			}
+			filteredDrainToDelete = append(filteredDrainToDelete, bucket)
+		}
+		drainToDelete = filteredDrainToDelete
 	}
 
 	if len(drainToDelete) > 0 {
@@ -254,7 +284,7 @@ func (a *Actuator) taintNodesSync(ctx context.Context, nodeGroupViews []*budgets
 	// Taint nodes concurrently and record failures
 	failedNodesChan := make(chan taintResult, len(nodesToTaint))
 	failedNodes := make(map[types.UID]struct{})
-	workqueue.ParallelizeUntil(context.Background(), maxConcurrentNodesTainting, len(nodesToTaint), func(piece int) {
+	workqueue.ParallelizeUntil(ctx, maxConcurrentNodesTainting, len(nodesToTaint), func(piece int) {
 		node := nodesToTaint[piece]
 		err := a.taintNode(ctx, node)
 		if err != nil {
@@ -267,7 +297,7 @@ func (a *Actuator) taintNodesSync(ctx context.Context, nodeGroupViews []*budgets
 		a.autoscalingCtx.Recorder.Eventf(result.node, apiv1.EventTypeWarning, "ScaleDownFailed", "failed to mark the node as toBeDeleted/unschedulable: %v", result.err)
 	}
 
-	// Parial taint actualtion disabled: failure mode
+	// Partial taint actuation disabled: failure mode
 	// Clean up all applied taints.
 	if len(failedNodes) > 0 && !a.autoscalingCtx.AutoscalingOptions.PartialTaintActuationEnabled {
 		// Clean up already applied taints in case of issues.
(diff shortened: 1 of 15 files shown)


pkg/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -61,21 +61,27 @@ func NewUpdateLatencyTracker(nodeLister kubernetes.NodeLister) *UpdateLatencyTra
 		finishTimestamp:          map[string]time.Time{},
 		remainingNodeCount:       0,
 		nodeLister:               nodeLister,
-		StartTimeChan:            make(chan nodeTaintStartTime),
+		StartTimeChan:            make(chan nodeTaintStartTime, 10000),
```
damikag: Why do we initialize this chan for a length of `10000`? I am wondering why we are over-provisioning here. If there is a valid reason still it would be better to extract this to a const.

pkg/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -25,7 +25,8 @@ import (
 )
 
 const sleepDurationWhenPolling = 50 * time.Millisecond
-const waitForTaintingTimeoutDuration = 30 * time.Second
+
+var waitForTaintingTimeoutDuration = 30 * time.Second
```
damikag, line 30: Shouldn't this be const?

pkg/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -42,16 +43,15 @@ type UpdateLatencyTracker struct {
 	// Sends node tainting start timestamps to the tracker
 	StartTimeChan            chan nodeTaintStartTime
 	sleepDurationWhenPolling time.Duration
-	// Passing a bool will wait for all the started nodes to get tainted and calculate
-	// latency based on latencies observed. (If all the nodes did not get tained within
-	// waitForTaintingTimeoutDuration after passing a bool, latency calculation will be
-	// aborted and the ResultChan will be closed without returning a value) Closing the
-	// AwaitOrStopChan without passing any bool will abort the latency calculation.
-	AwaitOrStopChan chan bool
+	// ExpectedNodeCountChan receives the capacity limit seeded precisely to the count
```
damikag: You meant "needed"?

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -207,57 +226,191 @@ func (a *Actuator) taintNodesSync(NodeGroupViews []*budgets.NodeGroupView) (time
...
-			a.pastLatencies.RegisterElement(latency)
-			a.pastLatencies.DropNotNewerThan(time.Now().Add(-1 * pastLatencyExpireDuration))
-			// CA is expected to wait 3 times the round-trip time between CA and the api-server.
-			// At this point, we have already tainted all the nodes.
-			// Therefore, the nodeDeleteDelayAfterTaint is set 2 times the maximum latency observed during the last hour.
-			nodeDeleteDelayAfterTaint = 2 * maxLatency(a.pastLatencies.ToSlice())
+// cleanNodesToBeDeleted executes the rollback of ToBeDeleted taints concurrently across all nodes provided.
+// This is typically deferred to run at the absolute end of the StartDeletion loop so as not to stall the healthy path.
+func (a *Actuator) cleanNodesToBeDeleted(nodes []*apiv1.Node) {
+	if a.autoscalingCtx.AutoscalingOptions.PartialTaintActuationEnabled {
```
damikag: Why do we branch `cleanNodesToBeDeleted()` based on `a.autoscalingCtx.AutoscalingOptions.PartialTaintActuationEnabled`?
It would be much cleaner do the same for both cases.

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -207,57 +226,191 @@ func (a *Actuator) taintNodesSync(NodeGroupViews []*budgets.NodeGroupView) (time
...
+					nodeDeleteDelayAfterTaint = 2 * maxLatency(a.pastLatencies.ToSlice())
+				}
+			}
+		}
+
+		if len(failedNodes) > 0 {
+			klog.Warningf("Couldn't taint %d nodes with ToBeDeleted, proceeding with partial scale down or bucket cleanup", len(failedNodes))
+			for _, node := range nodesToTaint {
+				if err, ok := failedNodes[node.Name]; ok {
+					a.autoscalingCtx.Recorder.Eventf(node, apiv1.EventTypeWarning, "ScaleDownFailed", "failed to mark the node as toBeDeleted/unschedulable: %v", err)
```
damikag: Isn't the same event is already emitted inside `func (a *Actuator) taintNode(node *apiv1.Node) error`?

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -207,57 +226,191 @@ func (a *Actuator) taintNodesSync(NodeGroupViews []*budgets.NodeGroupView) (time
...
 			nodesToTaint = append(nodesToTaint, node)
 		}
 	}
-	failedTaintedNodes := make(chan struct {
+
+	var successfulNodeGroupViews []*budgets.NodeGroupView
+	var globalNodesToClean []*apiv1.Node
+	var retErr errors.AutoscalerError
+
+	if a.autoscalingCtx.AutoscalingOptions.PartialTaintActuationEnabled {
```
damikag: I don't think this is the best way to add  partial taint actuation. This could be integrated to the existing code more closely. As of now new logic is inside a if block for a.autoscalingCtx.AutoscalingOptions.PartialTaintActuationEnabled and the old logic is inside the else block. There are a lot of repetitions. Instead of this you could unify most of the code and decide what to do with the tainting result based on whether `PartialTaintActuationEnabled` is true or false. That would be more cleaner and readable.

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -207,57 +226,191 @@ func (a *Actuator) taintNodesSync(NodeGroupViews []*budgets.NodeGroupView) (time
...
+			isAtomic = true
+		} else if opts != nil && opts.ZeroOrMaxNodeScaling {
+			isAtomic = true
+		}
+
+		var successfulBucket *budgets.NodeGroupView
+		var toClean []*apiv1.Node
+
+		if isAtomic {
+			successfulBucket, toClean = resolveAtomicBucketFailures(bucket, failedNodes)
```
damikag: Instead of having two function for `resolveAtomicBucketFailures()` and `resolveNonAtomicBucketFailures()` you could have one function splitting the `bucket` into succeeded and failed tainting attempts.  Based of `isAtomic` you could then decided how to proceed. That way it would be more readable and easy to understand.

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -207,57 +226,191 @@ func (a *Actuator) taintNodesSync(NodeGroupViews []*budgets.NodeGroupView) (time
...
+	return successfulNodeGroupViews, globalNodesToClean
+}
 
-	if a.autoscalingCtx.AutoscalingOptions.DynamicNodeDeleteDelayAfterTaintEnabled {
-		updateLatencyTracker.ExpectedNodeCountChan <- len(nodesToTaint)
-		latency, ok := <-updateLatencyTracker.ResultChan
-		if ok {
-			a.pastLatencies.RegisterElement(latency)
-			a.pastLatencies.DropNotNewerThan(time.Now().Add(-1 * pastLatencyExpireDuration))
-			// CA is expected to wait 3 times the round-trip time between CA and the api-server.
```
damikag, line 255: This is a very important comment describing the dynamic node deletion delay feature. Please do not remove it.

pkg/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -96,17 +96,20 @@ func (u *UpdateLatencyTracker) Start() {
...
-		case ntst := <-u.StartTimeChan:
+		case ntst, ok := <-u.StartTimeChan:
+			if !ok {
+				break DrainLoop
+			}
 			u.startTimestamp[ntst.nodeName] = ntst.startTime
 		default:
-			return // Channel is empty, safely exit without blocking.
+			// Channel is empty, safely exit without blocking.
+			break DrainLoop
```
damikag: You can achieve the same thing with `return`.

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -551,11 +545,18 @@ func (a *Actuator) scaleDownNodeToReport(node *apiv1.Node, drain bool) (*status.
...
+func (a *Actuator) taintNode(node *apiv1.Node, tracker *UpdateLatencyTracker) error {
 	if _, err := taints.MarkToBeDeleted(node, a.autoscalingCtx.ClientSet, a.autoscalingCtx.CordonNodeBeforeTerminate); err != nil {
 		a.autoscalingCtx.Recorder.Eventf(node, apiv1.EventTypeWarning, "ScaleDownFailed", "failed to mark the node as toBeDeleted/unschedulable: %v", err)
 		return errors.ToAutoscalerError(errors.ApiCallError, err)
 	}
+	
+	if tracker != nil {
+		// Pushing the timestamp *after* the API call completes excludes any 
+		// delay caused by the client-side throttling (which is controlled by --kube-api-qps and --kube-api-burst flags). 
+		tracker.StartTimeChan <- nodeTaintStartTime{node.Name, time.Now()}
```
damikag: I believe this is not correct. Now you start measuring latency after the API call finished.

BigDarkClown:
@tetianakh please address feedback from @damikag, I'll double check the PR once you do :)

pkg/core/scaledown/actuation/update_latency_tracker.go
tetianakh: yes! thanks for noticing, it's a leftover from changes I rolled back

pkg/core/scaledown/actuation/actuator.go
tetianakh: We're measuring latency of the taint propagation - how long it takes for our informers to see the taint after the api call is finished. We need to exclude the client-side throttling from the latency somehow, and the throttling is deep inside the client library. We can add a constant factor of e.g. 500ms to account for the apiserver latency somehow. But currently because of the throttling we can wait 20s for a request to be sent, which leads to unreasonably long sleep times.

pkg/core/scaledown/actuation/actuator.go
tetianakh: Because at HEAD we clean up the taints one by one, and with the experiment enabled we clean them up concurrently.

pkg/core/scaledown/actuation/actuator.go
tetianakh: Correct me if I'm wrong but the comment is not factually correct. We're not waiting for the round trip time between CA and apiserver. Round trip means the time it takes to send a request and receive a response. We're waiting the time it takes to send the request, receive the response, and for the taint to propagate to our informers. And it's the last part that takes the most time.

tetianakh:
@damikag I removed the last commit from this PR (where I changed what we track as latency) and I'll open a separate PR once this one is merged.

pkg/core/scaledown/actuation/actuator.go
tetianakh: oh wait here we don't need the if/else at all, you're right

pkg/core/scaledown/actuation/update_latency_tracker.go
tetianakh: changed to the number of nodes in the batch

pkg/core/scaledown/actuation/actuator.go
tetianakh: refactored the whole function. PTAL

pkg/core/scaledown/actuation/actuator.go
tetianakh: I'm not sure if I understand what you mean, but I refactored the code to be hopefully a bit more readable

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -135,33 +136,79 @@ func (a *Actuator) startDeletion(ctx context.Context, empty, drain []*apiv1.Node
...
 		return status.ScaleDownNoNodeDeleted, nil, nil
 	}
 
+	abortedNodeGroups := make(map[string]bool)
+
 	if len(emptyToDelete) > 0 {
 		// Taint all empty nodes synchronously
-		nodeDeleteDelayAfterTaint, err := a.taintNodesSync(ctx, emptyToDelete)
+		taintRes, err := a.taintNodesSync(ctx, emptyToDelete)
+		if len(taintRes.nodesToClean) > 0 {
```
damikag, line 160: This check is redundant.

You could `nodesToClean = append(nodesToClean, taintRes.nodesToClean...)`

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -135,33 +136,79 @@ func (a *Actuator) startDeletion(ctx context.Context, empty, drain []*apiv1.Node
...
 		if err != nil {
 			return status.ScaleDownError, scaledDownNodes, err
 		}
 
-		emptyScaledDown := a.deleteAsyncEmpty(ctx, emptyToDelete, nodeDeleteDelayAfterTaint, force)
+		emptyScaledDown := a.deleteAsyncEmpty(ctx, taintRes.successfulNodes, taintRes.delayAfterTaint, force)
 		scaledDownNodes = append(scaledDownNodes, emptyScaledDown...)
 	}
 
+	if len(drainToDelete) > 0 {
```
damikag, line 188: This `if` adds redundancy(There is the same `if` condition right after this `if` condition) although it could crop `drainToDelete`. You could delegate the filter out part to a separate function.

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -194,71 +241,177 @@ func (a *Actuator) deleteAsyncEmpty(ctx context.Context, NodeGroupViews []*budge
 	return reportedSDNodes
 }
 
-// taintNodesSync synchronously taints all provided nodes with NoSchedule. If tainting fails for any of the nodes, already
-// applied taints are cleaned up.
-func (a *Actuator) taintNodesSync(ctx context.Context, NodeGroupViews []*budgets.NodeGroupView) (time.Duration, errors.AutoscalerError) {
+type taintNodesResult struct {
```
damikag, line 244: nit: `nodeTaintResult` sounds better. `taintResult` is even better since it removes the redundant "node" term.

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -194,71 +241,177 @@ func (a *Actuator) deleteAsyncEmpty(ctx context.Context, NodeGroupViews []*budge
...
+			}
 			_, _ = taints.CleanToBeDeleted(ctx, taintedNode, a.autoscalingCtx.ClientSet, a.autoscalingCtx.CordonNodeBeforeTerminate)
 		}
+		// No need to record taint propagation latency, all taints are cleaned up.
 		if a.autoscalingCtx.AutoscalingOptions.DynamicNodeDeleteDelayAfterTaintEnabled {
-			close(updateLatencyTracker.AwaitOrStopChan)
+			close(updateLatencyTracker.ExpectedNodeCountChan)
 		}
-		return nodeDeleteDelayAfterTaint, errors.NewAutoscalerErrorf(errors.ApiCallError, "couldn't taint %d nodes with ToBeDeleted", len(failedTaintedNodes))
+		return taintNodesResult{delayAfterTaint: nodeDeleteDelayAfterTaint, successfulNodes: nil, nodesToClean: nil}, errors.NewAutoscalerErrorf(errors.ApiCallError, "couldn't taint %d nodes with ToBeDeleted", len(failedNodes))
```
damikag, line 314: Why `successfulNodes: nil`? Isn't this leaves tainted node forever without going to deletion phase?

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -194,71 +241,177 @@ func (a *Actuator) deleteAsyncEmpty(ctx context.Context, NodeGroupViews []*budge
...
 	}
 
+	// Compute taint propagation latency for successfully tainted nodes
```
damikag, line 317: The way how those `if` blocks are introduced makes it less readable. I would refactor like this

```
if len(failedNodes) == 0 {
 ...
return
}

if a.autoscalingCtx.AutoscalingOptions.PartialTaintActuationEnabled {
 ...
return
}

// existing logic when a.autoscalingCtx.AutoscalingOptions.PartialTaintActuationEnabled == false
return
```

pkg/core/scaledown/actuation/actuator.go
damikag: As I remember this is exactly what we want to measure here. Once we taint a node, we wanted to measure how long does it takes for CA to observe the change.

damikag: Apparently https://github.com/kubernetes-sigs/cluster-autoscaler/commit/eae3fca09c924fff22733f285509617484c41deb broke the initial implementation and it could be one of the main reason for the high latency measurement that you are pointing out. Let's address this in a separate PR.

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -194,71 +241,177 @@ func (a *Actuator) deleteAsyncEmpty(ctx context.Context, NodeGroupViews []*budge
...
+		failedNodes[result.node.UID] = struct{}{}
+		a.autoscalingCtx.Recorder.Eventf(result.node, apiv1.EventTypeWarning, "ScaleDownFailed", "failed to mark the node as toBeDeleted/unschedulable: %v", result.err)
+	}
+
+	// Partial taint actuation disabled: failure mode
+	// Clean up all applied taints.
+	if len(failedNodes) > 0 && !a.autoscalingCtx.AutoscalingOptions.PartialTaintActuationEnabled {
 		// Clean up already applied taints in case of issues.
-		for taintedNode := range taintedNodes {
+		for _, taintedNode := range nodesToTaint {
```
damikag, line 304: Please help me to understand what benefit do we get by changing the existing implementation.

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -135,33 +136,79 @@ func (a *Actuator) startDeletion(ctx context.Context, empty, drain []*apiv1.Node
...
 	if len(emptyToDelete) == 0 && len(drainToDelete) == 0 {
 		return status.ScaleDownNoNodeDeleted, nil, nil
 	}
 
+	abortedNodeGroups := make(map[string]bool)
+
 	if len(emptyToDelete) > 0 {
 		// Taint all empty nodes synchronously
-		nodeDeleteDelayAfterTaint, err := a.taintNodesSync(ctx, emptyToDelete)
+		taintRes, err := a.taintNodesSync(ctx, emptyToDelete)
```
damikag, line 159: What do you think of adding `failedAtomicNodeGroups` map to the result? This way you can reduce the complexity added to the `(a *Actuator) startDeletion()`
With that you can easily filter out failed atomic nodes group from `drainToDelete` and also later do `taints.CleanToBeDeleted()` 

You can get rid of `nodesToClean`. 
If `PartialTaintActuationEnabled` is disabled, you can keep cleaning as how it is implemented before.
When `PartialTaintActuationEnabled` is enabled, if the nodes are from atomic node groups you can clean all the nodes from that group right away and report that group. If this happed for empty nodes we can simply filter out impacted node groups.
If this happed when tainting nodes that require draining, you have to check whether there are some nodes in the empty group that needs to be cleaned. With node group id you can check that and later clean.

Further you can simplify `(a *Actuator) resolveBucketFailures()` and change it to a function that sort nodes within the same group based on whether tainting succeed or failed.

pkg/core/scaledown/actuation/actuator.go
```diff
@@ -135,33 +136,79 @@ func (a *Actuator) startDeletion(ctx context.Context, empty, drain []*apiv1.Node
...
+		// If an original bucket didn't make it to successfulNodes, it was aborted.
+		// If the group is atomic, we must abort its siblings in the drain phase.
+		for _, bucket := range emptyToDelete {
+			if !successfulGroups[bucket.Group.Id()] {
+				opts, err := bucket.Group.GetOptions(ctx, a.autoscalingCtx.NodeGroupDefaults)
+				if err != nil || (opts != nil && opts.ZeroOrMaxNodeScaling) {
+					abortedNodeGroups[bucket.Group.Id()] = true
+				}
+			}
+		}
```
damikag, line 178: This part could be discarded if you incorporated the above suggestion to add failedAtomicNodeGroups to taint results. If not please extract this to a separate function to improve readability.

pkg/core/scaledown/actuation/update_latency_tracker.go
damikag: It seems this is not rolled back still.

PR #21, opened 2026-08-03
ci: remove global gc flag from workflow

viktoriaas:
#### What type of PR is this?

/kind failing-test

This change removes the global `-gc` flag from benchmarking CI workflow and moves default activation directly into the specific benchmark to prepare the CI for upcoming changes in PR #18 . This way, the existing benchmark retains intended `gc` behaviour without applying the flag to other benchmark suites in the package and causing CI tests to fail. The change is backward-compatible and independent from #18.

### DIFF
--- a/.github/workflows/ca-benchmark.yaml
+++ b/.github/workflows/ca-benchmark.yaml
@@ -39,13 +39,13 @@ jobs:
         shell: bash
         run: |
           cd base/
-          go test -v -bench=BenchmarkRunOnce -benchmem -count=8 ./pkg/core/bench/... -args -gc true | tee ../base.txt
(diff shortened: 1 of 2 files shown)


Choraden:
LGTM! 

One nit: could you improve the commit message to express why it's better to change the default there? Thanks!

Choraden:
/lgtm
/assign BigDarkClown
Please approve github workflows first to check if benchmark works correctly.

BigDarkClown:
```
Verifying hack/../hack/verify-gofmt.sh
Please run hack/update-gofmt.sh to fix the following files:
./pkg/main.go
```

I think this is an rebase issue. @viktoriaas, Could you rebase to current main branch?

PR #22, opened 2026-08-03
Support ResourceClaimTemplates in check-capacity ProvisioningRequests

### LINKED ISSUE #10065: Support ResourceClaimTemplates in check-capacity ProvisioningRequests

/area cluster-autoscaler
/area core-autoscaler
/wg device-management

CA expands ProvisioningRequest PodTemplates into [virtual Pods created only in memory](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/provisioningrequest/pods/pods.go).

For a real Pod referencing a ResourceClaimTemplate (RCT), Kubernetes creates a unique [ResourceClaim (RC)](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/resourceclaim/controller.go) & records the RC name in Pod status. However, this never happens for virtual Pods because they're not actually created in the API server.

Currently, CA's [DRA snapshot](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/simulator/dynamicresources/snapshot/snapshot.go) can only resolve a direct RC reference or an RC name from Pod status. A virtual Pod with an RCT has neither. DRA scheduling therefore fails with `ResourceClaim not created yet` and the Pod is treated as unschedulable.

As such, `check-capacity` always reports `CapacityIsNotFound` for ProvisioningRequests containing Pods that reference an RCT, even when sufficient devices exist. This makes ProvisioningRequests unusable for admission decisions on DRA workloads referencing RCTs, since they are rejected as capacity not found regardless of actual cluster capacity.

When expanding a ProvisioningRequest, CA should materialize each RCT into an in-memory RC for each virtual Pod, without creating real RC objects in the cluster.

We considered having the ProvisioningRequest producer create real RCs and rewrite the YAML of each PodTemplate to reference those RCs directly. This does work, but it duplicates Kubernetes controller logic & creates API objects used only for simulation.

**Additional context.**:

Related issues:

- [#7686](https://github.com/kubernetes/autoscaler/issues/7686)
- [#7684](https://github.com/kubernetes/autoscaler/issues/7684)

Sid1279:
#### What type of PR is this?

/kind feature
/area cluster-autoscaler
/area core-autoscaler
/wg device-management

Cluster Autoscaler expands ProvisioningRequest PodTemplates into virtual Pods that exist only in memory. When a virtual Pod references a ResourceClaimTemplate, Kubernetes's ResourceClaim controller never observes it, so no ResourceClaim is created and no concrete claim name is added to the Pod's status.

The DRA snapshot therefore cannot resolve the claim, and `check-capacity` could incorrectly report `CapacityIsNotFound` even when sufficient capacity is available.

The code in this PR:
- materializes the ResourceClaims (in-memory) from referenced ResourceClaimTemplates 
- adds concrete ResourceClaim mappings onto the virtual Pods
- adds the virtual Pods and ResourceClaims to the same DRA snapshot transaction
- reuses the same workload construction for the initial capacity check and capacity booking subsequently
- supports both batch and non-batch check-capacity processing
- prevents the default non-batch path from including unrelated real pending Pods in the selected ProvisioningRequest
- reverts partially modified snapshot state when booking fails

The generated ResourceClaims are used only for scheduling simulation and are never created in the Kubernetes API.

A note on scope: ResourceClaimTemplate materialization in this PR is limited to virtual Pods generated from `check-capacity` ProvisioningRequests. This happens during the initial capacity check and when CA books that capacity for the request afterwards. It does not add ResourceClaimTemplate materialization for CapacityBuffer virtual Pods, and also does not add/change PVC handling for virtual Pods.

Fixes https://github.com/kubernetes/autoscaler/issues/10065

Related context:
- https://github.com/kubernetes/autoscaler/issues/7686
- https://github.com/kubernetes/autoscaler/pull/7530
- https://github.com/kubernetes/autoscaler/issues/7321

This PR fixes only the `check-capacity` ProvisioningRequest ResourceClaimTemplate path described in [#10065](https://github.com/kubernetes/autoscaler/issues/10065). The unschedulable-Pod snapshot refactor in [#7686](https://github.com/kubernetes/autoscaler/issues/7686), the ResourceClaimTemplate materialization for CapacityBuffer-generated virtual Pods and the PVC/generic-ephemeral-volume problem in [#7321](https://github.com/kubernetes/autoscaler/issues/7321) remain out of scope for this PR.

CA creates the simulated ResourceClaims only in memory for the initial capacity check, then recreates them with the same names/UIDs when booking the request's capacity in later autoscaler loops.

Deployments with `--enable-provisioning-requests=true` need `list` & `watch` permissions for `resource.k8s.io/resourceclaimtemplates`.

My testing:
- Unit tests and race tests
- Full `make test-ci` test suite
- Repository verification scripts, including `hack/verify-all.sh -v`
- End-to-end testing on a Kubernetes v1.35 cluster with NVIDIA DRA:
  - I created a real pending Pod with a ResourceClaimTemplate and:
     - confirmed it had a controller-generated ResourceClaim mapping
     - verified that the mapping remained unchanged while a separate ProvisioningRequest passed `check-capacity`
  - I submitted a separate workload with a ResourceClaimTemplate and saw:
    - its ProvisioningRequest passed `check-capacity`
    - the workload was admitted 
    - the workload's Pod received a real device allocation
- While the admitted workload Pod was active, I verified that:
    - exactly two ResourceClaims existed: 
      - one for the separate pending Pod
      - one for the admitted workload Pod
    - CA did not create any additional ResourceClaim in the cluster during this capacity simulation

```release-note
action required: check-capacity ProvisioningRequests now support Pods that reference ResourceClaimTemplates. Cluster Autoscaler materializes the implied ResourceClaims in memory only. Deployments using --enable-provisioning-requests must grant CA list & watch access to resource.k8s.io/resourceclaimtemplates.
```

```docs
- [Usage]: https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/
- [Other doc]: https://github.com/kubernetes/autoscaler/issues/10065
- [Other doc]: https://github.com/kubernetes/autoscaler/pull/7530
```

### DIFF
--- a/pkg/processors/provreq/processor.go
+++ b/pkg/processors/provreq/processor.go
@@ -24,6 +24,7 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	v1 "k8s.io/autoscaler/cluster-autoscaler/apis/provisioningrequest/autoscaling.x-k8s.io/v1"
 	"k8s.io/klog/v2"
 	ca_context "sigs.k8s.io/cluster-autoscaler/pkg/context"
@@ -55,11 +56,12 @@ type provReqProcessor struct {
 	client                         *provreqclient.ProvisioningRequestClient
 	injector                       injector
 	checkCapacityProcessorInstance string
+	simulationWorkloadBuilder      *provreq_pods.SimulationWorkloadBuilder
 }
 
 // NewProvReqProcessor return ProvisioningRequestProcessor.
-func NewProvReqProcessor(client *provreqclient.ProvisioningRequestClient, checkCapacityProcessorInstance string) *provReqProcessor {
-	return &provReqProcessor{now: time.Now, maxUpdated: defaultMaxUpdated, client: client, injector: scheduling.NewHintingSimulator(), checkCapacityProcessorInstance: checkCapacityProcessorInstance}
+func NewProvReqProcessor(client *provreqclient.ProvisioningRequestClient, checkCapacityProcessorInstance string, simulationWorkloadBuilder *provreq_pods.SimulationWorkloadBuilder) *provReqProcessor {
+	return &provReqProcessor{now: time.Now, maxUpdated: defaultMaxUpdated, client: client, injector: scheduling.NewHintingSimulator(), checkCapacityProcessorInstance: checkCapacityProcessorInstance, simulationWorkloadBuilder: simulationWorkloadBuilder}
 }
 
 // Refresh implements loop.Observer interface and will be run at the start
@@ -142,11 +144,20 @@ func (p *provReqProcessor) Process(ctx context.Context, autoscalingCtx *ca_conte
 // bookCapacity schedule fake pods for ProvisioningRequest that should have reserved capacity
 // in the cluster.
 func (p *provReqProcessor) bookCapacity(ctx context.Context, autoscalingCtx *ca_context.AutoscalingContext) error {
-	logger := klog.FromContext(ctx)
 	provReqs, err := p.client.ProvisioningRequests(ctx)
 	if err != nil {
 		return fmt.Errorf("couldn't fetch ProvisioningRequests in the cluster: %v", err)
 	}
+	return p.bookProvisioningRequests(ctx, autoscalingCtx, provReqs)
+}
+
+type capacityBooking struct {
+	provReq  *provreqwrapper.ProvisioningRequest
+	workload *provreq_pods.SimulationWorkload
+}
+
+func (p *provReqProcessor) bookProvisioningRequests(ctx context.Context, autoscalingCtx *ca_context.AutoscalingContext, provReqs []*provreqwrapper.ProvisioningRequest) error {
+	logger := klog.FromContext(ctx)
 	nodeInfos, err := autoscalingCtx.ClusterSnapshot.ListNodeInfos()
 	if err != nil {
 		return fmt.Errorf("couldn't list nodes: %v", err)
@@ -160,7 +171,7 @@ func (p *provReqProcessor) bookCapacity(ctx context.Context, autoscalingCtx *ca_
 			}
 		}
 	}
-	podsToCreate := []*apiv1.Pod{}
+	var bookings []capacityBooking
 	for _, provReq := range provReqs {
 		if !conditions.ShouldCapacityBeBooked(ctx, provReq, p.checkCapacityProcessorInstance) {
 			continue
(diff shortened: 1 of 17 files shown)


thc1006:
@Sid1279 — I opened #32 for this `findPodClaims` sparse-slice panic before I noticed your PR already carries the identical fix (and a better test). Sorry for the overlap; you were here first.

One thing worth surfacing: it's a crash regression from #8090, live on release-1.33 through 1.36 (1.32 uses `append` and is fine), whereas this PR is a feature that presumably won't be backported. Would you and the maintainers prefer I keep #32 as a small standalone fix so it can be cherry-picked to those branches, or fold it in here and handle the backport from this PR? Happy either way — just flagging so we're not racing the same hunk.

thc1006:
#32 now has semantic coverage for both lenient skip branches and a PredicateSnapshot-level RemoveNodeInfo regression case across the basic and delta stores. Since it is a small standalone crash regression and a cleaner backport unit, I would prefer to land #32 first, then rebase this PR to drop the overlapping snapshot.go and generic snapshot test hunks while keeping the ProvisioningRequest coverage.

@mtrqq does that ordering work for you? Happy to do it the other way and close #32 if you would rather this PR carry the fix.

mtrqq:
@Sid1279 Thanks a lot for your contribution and apologies for a delay in review, will try to make a first pass this week

@thc1006 Good point, thanks for mentioning; I agree let's first merge #32 as it's a small targeted fix and will be much faster to review

Sid1279:
@mtrqq sounds great! i'll rebase my PR once #32 gets merged to main. thank you!

pkg/provisioningrequest/pods/simulation_workload.go
```diff
@@ -0,0 +1,153 @@
...
+		Claims: make([]*resourcev1.ResourceClaim, 0),
+	}
+
+	for _, pod := range pods {
+		for claimIndex := range pod.Spec.ResourceClaims {
+			podClaim := &pod.Spec.ResourceClaims[claimIndex]
+			templateName := podClaim.ResourceClaimTemplateName
+			if templateName == nil {
+				continue
+			}
```
mtrqq, line 94: Is it possible to have a dependency on a shared function taken from the controller `k8s.io/dynamic-resource-allocation/resourceclaim` which will create in-memory resource claim object?

If we'll go ahead with hard-rolling this logic by creating a proper status object and placing the right annotation here and not depend on the controller utilities explicitly that'll definitely work right for now. But in the long term that may proof to be fragile as there's no link between these components and it may be challenging to pinpoint this discrepancy in resource claims creation for fake when debugging CA

pkg/provisioningrequest/pods/simulation_workload.go
```diff
@@ -0,0 +1,153 @@
...
+				Spec: *template.Spec.Spec.DeepCopy(),
+			})
+		}
+		workload.Pods = append(workload.Pods, pod)
+	}
+
+	return workload, nil
+}
+
+func simulationResourceClaimName(pod *corev1.Pod, logicalClaimName string) string {
```
mtrqq, line 146: I think we definitely want to add documentation for this function to explain scenarios which are covered here and especially the logic around truncation, it took a while from me to understand the podNameLength calculation part

pkg/provisioningrequest/pods/simulation_workload.go
```diff
@@ -0,0 +1,153 @@
...
+	identity := fmt.Sprintf("%s\x00%s\x00%s\x00%s", pod.Namespace, pod.Name, pod.UID, logicalClaimName)
```
mtrqq: Is there any reason why are we concatenating all the elements in the string first and then hashing it?

```
sha := sha256.New()
sha.Write(...)
sha.Write(...)
sha.Write(...)
sha.Sum(nil)
```

Current option while does that in two lines - using "\x00" immediately raises concern on "why?". I would either change the separator or use a longer version of the same code and potentially move that to a separate small helper

pkg/simulator/dynamicresources/snapshot/snapshot.go
```diff
@@ -313,8 +313,8 @@ func (s *Snapshot) findPodClaims(pod *apiv1.Pod, ignoreNotTracked bool) ([]*reso
 		return nil, nil
 	}
 
-	result := make([]*resourceapi.ResourceClaim, len(pod.Spec.ResourceClaims))
-	for claimIndex, claimRef := range pod.Spec.ResourceClaims {
+	result := make([]*resourceapi.ResourceClaim, 0, len(pod.Spec.ResourceClaims))
```
mtrqq, line 313: Reminder: As this issue is already patched and merged in the context of https://github.com/kubernetes-sigs/cluster-autoscaler/pull/32, can we rebase on latest master and remove it from the delta?

pkg/provisioningrequest/pods/simulation_workload.go
mtrqq: Also that would be ideal if we can extend our code coverage of this function beyond a single test case, that can also help greatly when trying to understand it

mtrqq (changes requested):
@Sid1279 Apologies for the review taking some time, but I'm already half-way through the PR and should finish the full pass tomorrow.

pkg/simulator/dynamicresources/snapshot/snapshot.go
Sid1279: I've just rebased this PR on top of #32 since it got merged to main earlier today.

pkg/provisioningrequest/pods/simulation_workload.go
Sid1279: I actually checked the current `k8s.io/dynamic-resource-allocation/resourceclaim` package but it doesn't expose a helper/function to materialize a `ResourceClaim` from a `ResourceClaimTemplate`.

Looks like [this logic](https://github.com/kubernetes/kubernetes/blob/v1.37.0-rc.0/pkg/controller/resourceclaim/controller.go#L1058-L1155) is internal to the K8s controller and creates the object via the K8s API using `GenerateName`. We need the CA fix to have a deterministic Name and UID (in-memory) since the same claim name must be reconstructed across simulations, so the K8s controller path can't be used directly.

Perhaps I can make this relationship more clear in the code? Maybe also add a `resourceclaim.IsForPod` assertion to verify the simulated `ResourceClaim` is owned by the virtual Pod?

mtrqq: Ideal state from my perspective would be to contribute to resource claim controller in order for it to expose a shared function which can be used by both us and the controller itself, one of the arguments of which may be name of such resource claim - controller may randomly pick one while we would deterministically define it

I don't have a strong preference to wait for this to be available before we merge this PR, but if we do - we need to follow-up with such a change quickly to avoid divergences moving on

@towca do you have an opinion on this one, can we follow-up with this change after the merge or should this be a pre-work?

pkg/provisioningrequest/checkcapacity/provisioningclass.go
```diff
@@ -140,7 +147,8 @@ func (o *checkCapacityProvClass) getProvisioningRequestsAndPods(unschedulablePod
 		if len(prs) == 0 {
 			return nil, nil
 		}
-		return []provreq.ProvisioningRequestWithPods{{PrWrapper: prs[0], Pods: unschedulablePods}}, nil
+		workload, err := o.simulationWorkloadBuilder.ForProvisioningRequest(prs[0])
+		return []provreq.ProvisioningRequestWithPods{{PrWrapper: prs[0], Workload: workload, Err: err}}, nil
```
mtrqq: I don't like current approach with error handling for the provisioning request and the fact that it's very different from the way how we do that with batching enabled

1. In batching: we skip provisioning request if we failed when creating pods for it and add status condition indicating that for the user
2. Single PV processing mode we are storing the error in the provisioning request wrapper and do not reflect anything in its state

Stream-lining these two methods makes sense from my side, unless I don't see something. Can you elaborate on that?

mtrqq (changes requested):

pkg/provisioningrequest/pods/simulation_workload.go
Sid1279: Yeah that makes sense, I'll document the function & truncation logic and add more test cases for this function in my next revision.

pkg/provisioningrequest/pods/simulation_workload.go
Sid1279: Yes I agree, I'll move this into a helper function

pkg/provisioningrequest/checkcapacity/provisioningclass.go
Sid1279: The non-batch error isn't lost though, we carry it in `ProvisioningRequestWithPods.Err` and then checkCapacityBatch adds the FailedToCreatePods condition, which is persisted by updateRequests. The failure is shown in the ProvisioningRequest but just happens later in the code flow.

But yeah I agree this could be streamlined so that the non-batch path immediately marks & persists the request as failed + skips it before capacity checkong + no longer carries the `Err` field.

pkg/provisioningrequest/pods/simulation_workload.go
Sid1279: @mtrqq Please let me know what the consensus is, and I can go ahead with the implementation!

pkg/provisioningrequest/checkcapacity/provisioningclass.go
Sid1279: I've made the non-batch path handle simulation workload creation failures immediately by adding & persisting the FailedToCreatePods condition, then skipping the request before checking for capacity. 

I also removed the Err field from ProvisioningRequestWithPods and removed the later error-handling branch. I added a regression test to verify that the failure is persisted + request is skipped too (TestGetProvisioningRequestsAndPodsNonBatchMarksMaterializationErrorsFailed)

pkg/provisioningrequest/pods/simulation_workload.go
Sid1279: I created a new helper (simulationResourceClaimHash) that hashes length-prefixed identity fields, keeping their boundaries unambiguous. I also added tests covering each field and ambiguous inputs.

pkg/provisioningrequest/pods/simulation_workload.go
Sid1279: I documented the deterministic naming + proportional truncation logic in `simulation_workload.go`.

I added comments about min char retention, DNS-safe trimming, etc... and also increased test coverage in `simulation_workload_test.go` to test boundary lengths, truncation, DNS validity, deterministic output.

pkg/provisioningrequest/pods/simulation_workload.go
Sid1279: @mtrqq while we wait for @towca's input, I added a comment pointing to the Kubernetes controller logic this code replicates. I also added a test using `resourceclaim.IsForPod` to make sure each simulated claim is properly associated with its Pod.

I have not added the shared helper yet, since we're yet to hear @towca's thoughts.

pkg/provisioningrequest/checkcapacity/provisioningclass.go
thc1006: On the error-handling streamline: one concrete case worth making sure either shape covers. `TrySchedulePods` turns a context cancellation or deadline into a partial `Result` with a nil error (`handleContextError`), so this path can't currently tell an interrupted simulation apart from a real capacity result. The `len(Statuses)` check then reads a canceled run as a miss, and on the booking side the Pods that happened to schedule before cancellation get treated as the scheduled set.

Checking `ctx.Err()` alongside the result and reverting with a transient error would cover it. `UnprocessedPods` alone wouldn't be enough to key on, since `breakOnFailure` also leaves later Pods unprocessed on a genuine miss. You two know the intended contract better, so flagging in case it's useful.

Sid1279: Oh yes good catch - I added the ctx.Err() checks after TrySchedulePods in both paths. Any interrupted simulations will now rollback and return a transient error (instead of treating partial results as valid capacity results).

Added tests for check-capacity and book capacity.

pkg/processors/provreq/processor.go
```diff
@@ -142,11 +144,21 @@ func (p *provReqProcessor) Process(ctx context.Context, autoscalingCtx *ca_conte
...
 	provReqs, err := p.client.ProvisioningRequests(ctx)
 	if err != nil {
 		return fmt.Errorf("couldn't fetch ProvisioningRequests in the cluster: %v", err)
 	}
+	return p.bookProvisioningRequests(ctx, autoscalingCtx, provReqs)
+}
+
+type capacityBooking struct {
+	provReq  *provreqwrapper.ProvisioningRequest
+	pods     []*apiv1.Pod
```
mtrqq: Is there a reason why can't we drop pods field from here and always rely on SimulationWorkload? As it anyway carries the pods

pkg/processors/provreq/processor.go
```diff
@@ -173,29 +185,95 @@ func (p *provReqProcessor) bookCapacity(ctx context.Context, autoscalingCtx *ca_
...
 	}
-	// Scheduling the pods to reserve capacity for provisioning request.
-	if _, err = p.injector.TrySchedulePods(context.Background(), autoscalingCtx.ClusterSnapshot, podsToCreate, false, clustersnapshot.SchedulingOptions{}); err != nil {
+
+	// Schedule all booking Pods together so pod priority is applied globally.
+	autoscalingCtx.ClusterSnapshot.Fork()
+	validBookings := make([]capacityBooking, 0, len(bookings))
+	var pods []*apiv1.Pod
+	for _, booking := range bookings {
+		if booking.workload != nil {
```
mtrqq: If we assume that we'll always have pods represented through the simulation workload - we would need to replace this this check for resource claims

pkg/provisioningrequest/pods/simulation_workload.go
mtrqq: This looks way more understandable from my end, thanks for documenting and testing that more thoroughly!

mtrqq (changes requested):
Thanks a lot for the contribution! Added two more comments, but should be relatively quick to get them resolved.

I'm ready to add LGTM once we align on the next steps with resource claim controller integration

pkg/processors/provreq/processor.go
Sid1279: Yup that makes sense! I removed the pods field from capacityBooking. So now both check-capacity & ordinary capacity bookings carry their Pods through the SimulationWorkloads.

pkg/processors/provreq/processor.go
Sid1279: I agree, now the claim insertion & cleanup check `len(workload.Claims) > 0` and scheduling uses `workload.Pods`

Sid1279:
Hi @mtrqq please let me know whenever you align on next steps regarding the resource claim controller integration! Or do let me know if there's anything I can do on my end.

PR #23, opened 2026-08-04
Move and partially fix Makefile and related files 

Choraden:
#### What type of PR is this?
/kind cleanup

This PR moves Makefile and its related file to the root directory and makes them operational after the migration. 
It does not fixes everything and should be treated as a step towards better state. 
Things may (and most probably will) still change with follow up PRs. 

Fixes #

```release-note
NONE
```

```docs

```


jackfrancis (changes requested):
@Choraden now that we've moved Makefile out of pkg/ I think we need to remove that working-directory line here:

https://github.com/kubernetes-sigs/cluster-autoscaler/blob/main/.github/workflows/ca-test.yaml#L36

pkg/cloudprovider/azure/test/Makefile
```diff
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 REPO_ROOT:=$(shell git rev-parse --show-toplevel)
-CAS_ROOT:=$(REPO_ROOT)/cluster-autoscaler
```
jackfrancis, line 16: Why are we preserving `pkg/cloudprovider/azure` at all in the new structure?

Choraden: We decided to temporarily preserve gce and azure providers here in case we wanted to run prow e2e tests.
The ultimate goal is to remove them and switch to kwok/kind e2e tests.

Choraden:
> @Choraden now that we've moved Makefile out of pkg/ I think we need to remove that working-directory line here:
> 
> https://github.com/kubernetes-sigs/cluster-autoscaler/blob/main/.github/workflows/ca-test.yaml#L36

Good catch. I removed it.

jackfrancis (approved):
/lgtm
/approve

I assume there will be many more of these, so let's not worry about perfection at this point

PR #25, opened 2026-08-04
Refactor cluster autoscaler flags to be injected on-demand and not on import

mtrqq:
What type of PR is this?
/kind cleanup

What this PR does / why we need it:

* Move flags initialization from global space to an on-demand function to produce the autoscaling options;
* Add tests which verify that everything parses correctly and tests for edge cases in argument parsing;

This direction is on the same line as making cluster autoscaler a library in the long-run, current approach makes it infeasible for forks to properly extend and control their CLI arguments, with this patch - it's now possible without loosing anything else.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/config/flags/flags.go
+++ b/pkg/config/flags/flags.go
@@ -19,6 +19,7 @@ package flags
 import (
 	"flag"
 	"fmt"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


mtrqq:
Moved PR from https://github.com/kubernetes/autoscaler/pull/9708

BigDarkClown:
Some minor nits, but nothing that should stop the merge. Let's get this in before you need to rebase :P

/lgtm
/approve

PR #29, opened 2026-08-06
Move the RegisterScaleUp metric calls to NodeGroupChangeMetricsProducer.

shaikenov:
#### What type of PR is this?

/kind cleanup

This is a copy of a [PR](https://github.com/kubernetes/autoscaler/pull/9667) from the old repo. 

Fixes #

During the implementation I figured out that we do not have a single test which covers the `RegisterScaleUp` metric emission. So I decided to have commits as follows:
1. Introduced unit tests which will check that the metric calls did actually happen (apparently we had only one place in `scaleUpExecutor`). Ensure that the metric is correctly called in there.
2. Moved the metric generation into `NodeGroupChangeMetricsProducer` and adjusted a little bit the test to the new implementation - double checked it by checking that tests from point 1 are still passing.
That is the reason that you will see some lines introduced in commit 1 getting deleted in commit 2. 
3. Added the `TemplateNodeInfoRegistry` as a param of `NodeGroupChangeMetricsProducer` to use up-to-date labels in the metric calls. More info in [here](https://github.com/kubernetes/autoscaler/pull/9667#issuecomment-5141883508)

Also since now metric is emitted from a single place -`NodeGroupChangeMetricsProducer` - there is no need to pass `nodeInfos` around as we already have a logic for `gpu` and `dra` info retrieval in the `NodeGroupChangeMetricsProducer`

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/clusterstate/clusterstate_test.go
+++ b/pkg/clusterstate/clusterstate_test.go
@@ -1978,6 +1978,10 @@ type mockMetrics struct {
 	mock.Mock
 }
 
+func (m *mockMetrics) RegisterScaleUp(delta int, gpuResourceName, gpuType, draDriverName string) {
+	m.Called(delta, gpuResourceName, gpuType, draDriverName)
+}
+
 func (m *mockMetrics) RegisterFailedScaleUp(reason metrics.FailedScaleUpReason, gpuResourceName, gpuType, draDrivers string) {
 	m.Called(reason, gpuResourceName, gpuType, draDrivers)
 }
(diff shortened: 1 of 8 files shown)


shaikenov:
/area cluster-autoscaler

This is a rebased copy of an old PR from the old repo: https://github.com/kubernetes/autoscaler/pull/9667.

Thanks @Choraden, for your [note](https://github.com/kubernetes/autoscaler/pull/9667#issuecomment-5141883508) and suggestions about the possible regression in the metric labels. I addressed in here in the last commit. Could you PTAL if it is alright now?

/assign Choraden

Choraden:
@shaikenov thanks for the change! I'll once again say that I'm huge fan of that refactor. Removing nodeInfos from scaleUp params is a huge win. Adding a TemplateNodeInfoRegistry seems to be solving expensive TemplateNodeInfo() calls in the happy path.

/lgtm

nit: Please improve commit messages so that they express the motivation behind the change in more detail.

PR #32, opened 2026-08-06
Fix nil-pointer panic in DRA snapshot for untracked pod claims

### LINKED ISSUE #31: cluster-autoscaler panics (nil-pointer) when removing a node whose Pod references an untracked DRA ResourceClaim
/area cluster-autoscaler
/kind bug

cluster-autoscaler, specifically the in-memory DRA (dynamic resource allocation) cluster snapshot under `pkg/simulator/dynamicresources/snapshot`.

**What version of the component are you using?**:

Component version: `main` at [ea003e7](https://github.com/kubernetes-sigs/cluster-autoscaler/commit/ea003e7e16b46d005fb8c13ee24c95b857d2b469).

**What k8s version are you using (`kubectl version`)?**:

Not tied to a specific server version. The bug is in cluster-autoscaler's own snapshot bookkeeping and is exercised whenever CA runs with `--enable-dynamic-resource-allocation`. I reproduced it with a unit test rather than on a live cluster (recipe below).

**What environment is this in?**:

Any Cluster Autoscaler deployment with DRA enabled can be affected when node teardown encounters the owner/direct-consumer sequence described above: an earlier Pod on the node owns an auto-generated ResourceClaim and a later Pod on the same node references that claim directly.

**What did you expect to happen?**:

`RemovePodOwnedClaims` is explicitly documented to tolerate claims that are not in the snapshot: *"This method removes all relevant claims that are in the snapshot, and doesn't error out if any of the claims are missing"* ([snapshot.go#L137-L139](https://github.com/kubernetes-sigs/cluster-autoscaler/blob/ea003e7e16b46d005fb8c13ee24c95b857d2b469/pkg/simulator/dynamicresources/snapshot/snapshot.go#L137-L139)). So when `RemoveNodeInfo` tears down a node whose pods still reference a ResourceClaim that is no longer tracked, I expected that claim to be skipped and node removal to proceed.

**What happened instead?**:

cluster-autoscaler panics with a nil-pointer dereference inside [`GetClaimId`](https://github.com/kubernetes-sigs/cluster-autoscaler/blob/ea003e7e16b46d005fb8c13ee24c95b857d2b469/pkg/simulator/dynamicresources/snapshot/snapshot.go#L39), called from `RemovePodOwnedClaims`. The panic unwinds out of `RemoveNodeInfo` and takes down the autoscaler loop.

The cause is in [`findPodClaims`](https://github.com/kubernetes-sigs/cluster-autoscaler/blob/ea003e7e16b46d005fb8c13ee24c95b857d2b469/pkg/simulator/dynamicresources/snapshot/snapshot.go#L311-L351):

```go
result := make([]*resourceapi.ResourceClaim, len(pod.Spec.ResourceClaims))
for claimIndex, claimRef := range pod.Spec.ResourceClaims {
    claimName := claimRefToName(pod, claimRef)
    if claimName == "" {
        if !ignoreNotTracked { return nil, ... }
        continue                 // result[claimIndex] stays nil
    }
    claimId := ResourceClaimId{Name: claimName, Namespace: pod.Namespace}
    claim, found := s.resourceClaims.FindValue(claimId)
    if !found {
        if !ignoreNotTracked { return nil, ... }
        continue                 // result[claimIndex] stays nil
    }
    result[claimIndex] = claim
}
return result, nil
```

The slice is sized to the number of claim references and written by index. On the `ignoreNotTracked == true` path (only `RemovePodOwnedClaims` uses it) a skipped claim leaves a `nil` entry at its index. `RemovePodOwnedClaims` then ranges over the result and calls `GetClaimId(claim)` on every element, including the `nil` one, dereferencing a nil `*ResourceClaim`.

**How to reproduce it (as minimally and precisely as possible)**:

A unit test in the snapshot package: a Pod referencing a claim that is absent from the snapshot, run through the `RemovePodOwnedClaims` path.

```go
claims := map[ResourceClaimId]*resourceapi.ResourceClaim{
    GetClaimId(sharedClaim1):  sharedClaim1.DeepCopy(),
    GetClaimId(sharedClaim2):  sharedClaim2.DeepCopy(),
    GetClaimId(pod1OwnClaim1): pod1OwnClaim1.DeepCopy(),
    // pod1OwnClaim2, which pod1 references, is intentionally omitted
}
snapshot := NewSnapshot(claims, nil, nil, nil)
snapshot.RemovePodOwnedClaims(pod1)   // panics on current main
```

```
panic: runtime error: invalid memory address or nil pointer dereference
...snapshot.GetClaimId(...)
...snapshot.(*Snapshot).RemovePodOwnedClaims(...)
```

In production the same path is reached through [`RemoveNodeInfo`](https://github.com/kubernetes-sigs/cluster-autoscaler/blob/ea003e7e16b46d005fb8c13ee24c95b857d2b469/pkg/simulator/clustersnapshot/predicate/predicate_snapshot.go#L236-L260), which loops over the pods of the node being removed and calls `RemovePodOwnedClaims(pod.Pod)` for each. Note that the DRA snapshot is built once by `SetClusterState`, which strict-validates every scheduled pod's claims (`PodClaims`) before storing them, so a pod whose claim is missing at build time fails the build rather than being stored silently. The untracked-claim condition arises during teardown itself: `RemovePodOwnedClaims` deletes the claims a pod owns, so if an earlier pod on the node owns a claim that a later pod on the same node also references, the earlier pod's cleanup removes it and the later pod then hits the lenient missing-claim path — the nil hole — and panics. This is also why the method uses `ignoreNotTracked=true`: missing claims on this path are anticipated.

**Anything else we need to know?**:

The fix is a three-line change (build the result with `append`). PR #32 carries it with semantic unit coverage for both lenient skip branches and a `PredicateSnapshot`-level `RemoveNodeInfo` regression case across the basic and delta snapshot stores. This is a regression from the patch-set snapshot refactor (kubernetes/autoscaler#8090): the same fixed-length `findPodClaims` is present on release-1.33 through 1.36, while 1.32 is unaffected (it uses `append`). Note that #22 already carries the identical production fix as part of its ResourceClaimTemplate work; I've cross-linked there to coordinate which PR should carry it.

thc1006:
#### What type of PR is this?

/kind bug

`findPodClaims` in the DRA cluster snapshot builds its result slice sized to the number of the Pod's claim references and fills it by index. On the lenient path (`ignoreNotTracked == true`) a claim that has no resolved name, or that is not tracked in the snapshot, is skipped with `continue`, which leaves a `nil` hole at that index. `RemovePodOwnedClaims` is the only caller that uses that path, and it ranges over the returned slice calling `GetClaimId(claim)` on every element. The `nil` hole makes `GetClaimId` dereference a nil `*ResourceClaim`, and the autoscaler panics.

That contradicts the method's own contract: *"This method removes all relevant claims that are in the snapshot, and doesn't error out if any of the claims are missing."* A missing claim is meant to be skipped, not to crash the process.

It is reachable during node teardown. `RemoveNodeInfo` iterates the pods of the node being removed and calls `RemovePodOwnedClaims` for each. When an earlier Pod on the node owns an auto-generated ResourceClaim and a later Pod on the same node references that claim directly, cleanup of the owner removes the claim, so cleanup of the later Pod then finds the reference missing, which is enough to take the autoscaler loop down when CA runs with DRA enabled. Kubernetes permits a direct reference to an auto-generated claim, although it is discouraged because the claim's lifetime stays tied to the owner Pod.

The fix builds the result with a zero-length slice and `append`, so only claims that are actually found are added and there are no `nil` entries:

```go
result := make([]*resourceapi.ResourceClaim, 0, len(pod.Spec.ResourceClaims))
...
result = append(result, claim)
```

The strict callers (`PodClaims`, `ReservePodClaims`, `UnreservePodClaims`, all `ignoreNotTracked == false`) see no change. On that path any missing claim returns early with an error, and when nothing is missing every claim is appended in iteration order, so the returned slice has the same length and order as the old index-filled one. Only the lenient `RemovePodOwnedClaims` path is affected, and it drops the `nil` holes it was never supposed to produce. The consumers of the exported `PodClaims` bear this out: each checks the error first, then either ranges over the slice ([`verifyScheduledPodResourceClaims`](https://github.com/kubernetes-sigs/cluster-autoscaler/blob/ea003e7e16b46d005fb8c13ee24c95b857d2b469/pkg/simulator/clustersnapshot/predicate/predicate_snapshot.go#L449)) or hands it whole to `framework.NewPodInfo`; none index it positionally.

Fixes #31

Related: #22

The regression tests fail on the current code with `panic: ... nil pointer dereference ... GetClaimId` and pass with the fix. They drive the real `RemovePodOwnedClaims` path with a snapshot missing one of the Pod's referenced claims, assert the final claim state, and check that cleanup still reaches the claims after the skipped reference. One case covers a claim absent from the snapshot, the other an unresolved template name, so both `continue` branches in `findPodClaims` are exercised.

A `PredicateSnapshot`-level case also covers the concrete `RemoveNodeInfo` owner/direct-consumer teardown sequence from #31, and runs through the existing basic and delta snapshot stores with their fork, commit, and revert variants.

#22 currently carries the same `findPodClaims` change as part of its larger ResourceClaimTemplate work. Whichever lands first, the other can drop the overlapping hunk. I am happy to defer to whatever ordering you prefer, or to close this in favor of #22.

I verified `main` only. The same fixed-length implementation exists on the release-1.33 through release-1.36 branches; branch tests were not run, so backport eligibility and workflow are your call.

```release-note
Fixed a nil-pointer panic in cluster-autoscaler's DRA snapshot when removing a node whose Pod references a ResourceClaim that another Pod's cleanup had already removed from the snapshot. Node removal now skips the missing claim instead of crashing the autoscaler.
```

```docs

```

### DIFF
--- a/pkg/simulator/clustersnapshot/predicate/predicate_snapshot_test.go
+++ b/pkg/simulator/clustersnapshot/predicate/predicate_snapshot_test.go
@@ -766,8 +766,8 @@ func validTestCases(t *testing.T, snapshotName string) []modificationTestCase {
 			// Two pods on the same node: podWithClaims owns podOwnedClaim (and references sharedClaim),
 			// and podConsumesOwned references podOwnedClaim directly. Teardown removes the owner's claim
 			// while processing the first pod, so cleanup of the second pod must tolerate the now-missing
-			// reference instead of panicking. podOwnedClaim is reserved for both pods so SetClusterState
-			// accepts the state.
+			// reference instead of panicking. podOwnedClaim is allocated and reserved for both pods to
+			// model the scheduled state.
 			state: snapshotState{
 				nodes:       []*apiv1.Node{node},
 				csiSnapshot: createCSISnapshot(csiNode),
--- a/pkg/simulator/clustersnapshot/predicate/predicate_snapshot_test.go
+++ b/pkg/simulator/clustersnapshot/predicate/predicate_snapshot_test.go
@@ -784,7 +784,7 @@ func validTestCases(t *testing.T, snapshotName string) []modificationTestCase {
 				return snapshot.RemoveNodeInfo(node.Name)
 			},
 			// Both pods leave with the node. The pod-owned claim is removed, and the shared claim
-			// stays with the owner's reservation cleared. Before the fix this panics.
+			// stays with the owner's reservation cleared.
 			modifiedState: snapshotState{
 				draSnapshot: drasnapshot.NewSnapshot(
 					map[drasnapshot.ResourceClaimId]*resourceapi.ResourceClaim{
--- a/pkg/simulator/clustersnapshot/predicate/predicate_snapshot_test.go
+++ b/pkg/simulator/clustersnapshot/predicate/predicate_snapshot_test.go
@@ -262,6 +262,11 @@ func validTestCases(t *testing.T, snapshotName string) []modificationTestCase {
 	largePod := BuildTestPod("largePod", 999, 999)
 	podWithClaims := BuildTestPod("podWithClaims", 1, 1,
 		WithResourceClaim("claim1", "sharedClaim", ""), WithResourceClaim("claim2", "podOwnedClaim", "podOwnedClaimTemplate"))
+	// podConsumesOwned references podWithClaims' auto-generated podOwnedClaim directly, without
+	// owning it. When both pods are on the same node, teardown removes the owner's claim first, so
+	// cleanup of this pod then finds that reference missing.
+	podConsumesOwned := BuildTestPod("podConsumesOwned", 1, 1,
+		WithResourceClaim("claim1", "podOwnedClaim", ""))
 
 	podOwnedClaim := drautils.TestClaimWithPodOwnership(podWithClaims,
 		&resourceapi.ResourceClaim{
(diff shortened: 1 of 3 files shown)


thc1006:
Cross-linking: this `findPodClaims` change is byte-for-byte the same as a hunk already in #22, which predates this PR and pairs it with a stronger regression test (it asserts the post-cleanup claim set, not just the absence of a panic). I missed it in my dedup — apologies for the duplicate.

It is a crash regression from the patch-set snapshot refactor (kubernetes/autoscaler#8090), and the fixed-length `findPodClaims` is present on release-1.33 through 1.36 (1.32 is unaffected — it still uses `append`). Since #22 is a feature that likely won't be backported, a small standalone fix might be worth keeping just for cherry-picking — but I'm equally happy to close this in favour of #22. Deferring to @Sid1279 and the maintainers; I've left a note on #22 to coordinate.

pkg/simulator/clustersnapshot/predicate/predicate_snapshot_test.go
```diff
@@ -756,6 +761,37 @@ func validTestCases(t *testing.T, snapshotName string) []modificationTestCase {
...
+						drasnapshot.GetClaimId(sharedClaim):   drautils.TestClaimWithPodReservations(drautils.TestClaimWithAllocation(sharedClaim, sharedClaimAlloc), podWithClaims),
+					},
+					map[string][]*resourceapi.ResourceSlice{node.Name: resourceSlices},
+					nil, deviceClasses),
+			},
+			op: func(snapshot clustersnapshot.ClusterSnapshot) error {
+				return snapshot.RemoveNodeInfo(node.Name)
+			},
+			// Both pods leave with the node. The pod-owned claim is removed, and the shared claim
+			// stays with the owner's reservation cleared. Before the fix this panics.
```
mtrqq: nit: I would remove "Before the fix this panics." as it helps a lot during the review, but may be confusing once we merge this change

mtrqq:
Elegant fix, thanks for contribution! Added a single nit, once it's resolved - will add LGTM

mtrqq (changes requested):

pkg/simulator/clustersnapshot/predicate/predicate_snapshot_test.go
thc1006: Removed it, thanks!

towca:
The fix LGTM!

/approve

```
When an earlier Pod on the node owns an auto-generated ResourceClaim and a later Pod on the same node references that claim directly
```

This is something we definitely didn't account for when designing DRA integration for Cluster Autoscaler. Keep in mind that there might be more places in CA where we assume that "Pod-owned claims are only referenced by the owner". I'm really curious, could you share a use-case where such a setup happens? Is it one Pod with a claim template, and another Pod meant to be collocated with the first one - only created after the claim for the first one is actually rendered? What does this setup address?

PR #36, opened 2026-08-07
[feature]: DeleteNodes in batches instead of one by one

LucasAndFlores:
#### What type of PR is this?
/kind feature

This PR implements the logic to delete multiple nodes if the flag `node-deletion-batcher-interval` is set. If it is, the cloudprovider/clusterapi will try to delete multiple replicas instead of one by one.

None

```release-note
If the flag `node-deletion-batcher-interval` is set and the cloudprovider is clusterapi, the nodes will be deleted in batches instead of one by one
```

### DIFF
--- a/pkg/cloudprovider/clusterapi/clusterapi_controller.go
+++ b/pkg/cloudprovider/clusterapi/clusterapi_controller.go
@@ -22,6 +22,7 @@ import (
 	"path"
 	"strings"
 	"sync"
+	"time"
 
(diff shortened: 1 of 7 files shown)


LucasAndFlores:
We understood that this PR should target the old repo since we will migrate cluster-api to a new repo. I'm closing this one and I opened [this one](https://github.com/kubernetes/autoscaler/pull/10197) targeting the old repo.

PR #38, opened 2026-08-09
feat: DRA cluster limits

### LINKED ISSUE #8184: CA DRA: integrate with cluster-wide resource limits

/area cluster-autoscaler
/area core-autoscaler
/wg device-management

Cluster Autoscaler allows users to configure cluster-wide resource limits which are taken into account when making scaling decisions, together with the usual NodeGroup Node limits. The resource limits can be defined for standard resources (CPU, memory), as well as custom extended resources (e.g. `nvidia.com/gpu`). However, making the limits work with a custom extended resource requires writing a `CustomResourcesProcessor` which translates a `Node` to a list of custom resources (`<resource_type`, `resource_count>`).

We have a `CustomResourcesProcessor` which handles `nvidia.com/gpu` by asking CloudProvider for a "GPU type label", and taking the value of that label from the Node. So if we have a Node with allocatable `nvidia.com/gpu: 4` and a `cloud.provider.com/gpu-type: nvidia-tesla-t4` label, the processor will translate it into `nvidia-tesla-t4: 4`. Whenever CA wants to provision such a Node, it checks if there is a limit configured for `nvidia-tesla-t4`, and if so if it allows adding 4 more resources.

It's not clear how to extend this model for DRA resources, because DRA devices are expressed differently than extended resources.

Using DRA, Node-local devices are exposed as 3-tuples: `<driver_name, pool_name, device_name>`. Taking the example above, a Node with 4 `nvidia-tesla-t4` GPUs would expose 4 separate devices, e.g.: `<nvidia.com, name-of-the-node, gpu-0>, <nvidia.com, name-of-the-node, gpu-1>, <nvidia.com, name-of-the-node, gpu-2>, <nvidia.com, name-of-the-node, gpu-3>`. Each such device would also have attribute maps, which should contain the `nvidia-tesla-t4` value _somewhere_.

CA needs to somehow be able to figure out the "type" of a given device to configure and check limits on - by looking at the device attributes. We could follow the current example and ask CloudProvider for which attribute to look at, but IMO it'd be better to standardize a "device type" or a "device type for the purpose of autoscaling limits" parameter at the Kubernetes layer.

MenD32:
#### What type of PR is this?
/kind feature

cluster autoscaler currently does not have a limit mechanism that'd bound the amount of devices that can be scaled up / down.

Fixes https://github.com/kubernetes/autoscaler/issues/8184

```release-note
a new flag `dra-total` has been added to cluster-autoscaler entrypoint - this flag configures the bounds of DRA autscaling.
The flag is of the format "<driver>/<attribute-name>=<attribute-value>:<min>:<max>" - where driver is the name of the DRA driver, attribute name and attribute value reflect the DRA device attributes by which devices will be selected, and min and max set the lower and upper bounds for autoscaling, f.e.
- "gpu.nvidia.com/productName=A100:16:64" will stop a scaling operation if it will cause the cluster to contain more than 64 or less than 16 DRA device with the gpu.nvidia.com driver and with the productName attribute set to `A100`
```

```docs

```

### DIFF
--- a/pkg/config/autoscaling_options.go
+++ b/pkg/config/autoscaling_options.go
@@ -17,6 +17,7 @@ limitations under the License.
 package config
 
 import (
+	"fmt"
 	"time"
 
 	kubelet_config "k8s.io/kubernetes/pkg/kubelet/apis/config"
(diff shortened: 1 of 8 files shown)


MenD32:
@towca may I tag you as reviewer?

PR #39, opened 2026-08-10
Update Kubernetes dependencies to v1.37.0-rc.0

### LINKED ISSUE #10051: New NodeInfo and NodeInfoLister interfaces to support InterPodAffinityHostnameFastPath

/area cluster-autoscaler

https://github.com/kubernetes/kubernetes/pull/138198 introduces a new method on `k8s.io/kube-scheduler/framework.NodeInfoLister`:
- `HavePodsWithRequiredNonHostScopedAntiAffinityList() ([]NodeInfo, error)`

https://github.com/kubernetes/kubernetes/pull/138198 introduces a new method on `k8s.io/kube-scheduler/framework.NodeInfo`:
- `GetPodsWithRequiredNonHostScopedAntiAffinity() []PodInfo`

Both are backed by a new `NodeInfo.PodsWithRequiredNonHostScopedAntiAffinity` subset that
the scheduler populates in `AddPodInfo`/`RemovePod` **only when the `InterPodAffinityHostnameFastPath` feature gate is enabled** (empty otherwise).

Feature gate is disabled by default, so the bare minimum is implementing the interface to ensure compilation once the change lands (may land in 1.37).

We can also implement `InterPodAffinityHostnameFastPath` in CA to get binpacking perf benefits with the interface consumption, or do that as separate, follow-up work.

**Additional context.**:

mtrqq:
#### What type of PR is this?

/kind cleanup

This PR updates dependencies to use kubernetes 1.37.0-rc.0 release tag in preparation for the 1.37 release of autoscaler

In this chain we are addressing following incompatibilities:

1. [Implementation of KEP-6012](https://github.com/kubernetes/kubernetes/pull/139596/changes#diff-5b8b417b1f669c2c46b68aac7422f5501074601aa329c83949e96345b0e30305) and [scheduler caching change](https://github.com/kubernetes/kubernetes/pull/140077) added additional methods to scheduler interface to interact with pod groups. Right now cluster autoscaler doesn't support them, this PR adds no-op methods to fill the gaps in the interface
2. [InterPodAffinityHostnameFastPath feature](https://github.com/kubernetes/kubernetes/pull/138198) added new methods to scheduler API which were implemented and tested in the context of https://github.com/kubernetes-sigs/cluster-autoscaler/pull/17 and cherry-picked here to have a single change which stabilizes interface discrepancies (Thanks @tetianakh!)
3. NodeDeclaredFeatures tests which required disablement of the feature flag were adjusted to test it differently as the feature gate is locked to true and cannot be overriden
4. Env test binaries cannot be downloaded yet as there's no released 1.37 version yet, as part of this change we are adding defaulting layer which fallbacks to latest in case not able to acquire binaries for specific version

Fixes https://github.com/kubernetes/autoscaler/issues/10051

See https://github.com/kubernetes-sigs/cluster-autoscaler/pull/17 for the in-depth explanation of the anti-affinity optimization change

```release-note
Improved performance for interpod affinity and anti-affinity with topologyKey "kubernetes.io/hostname"
```

```docs

```


mtrqq:
@norbertcyran I see that there's no binaries for 1.37 to run env-tests, do we want to provide some fallback to other k8s versions or a way to avoid running these tests during new version release?

PR #40, opened 2026-08-10
Mark all pod groups unschedulable when no scale-up options are available

ttetyanka:
When prepareScaleUp finds no expansion options (len(options) == 0), mark the pod equivalence groups as unschedulable with NoScaleUpOptionsAvailableReason before returning the status. This ensures that unschedulable reasons are properly populated in the ScaleUpStatus.
#### What type of PR is this?
/kind bug
When `prepareScaleUp` finds no expansion options (`len(options) == 0`), it previously returned `noOptionsAvailableStatus` without marking the pod equivalence groups as unschedulable. As a result, the unschedulable reasons for remaining pods were not properly populated in the `ScaleUpStatus`.
This PR introduces `NoScaleUpOptionsAvailableReason` and marks all pod equivalence groups with this reason when no expansion options are available, ensuring accurate scale-up status reporting.
Fixes #
This aligns the `len(options) == 0` branch with the other early exits in `prepareScaleUp` (such as `ExpansionOptionsFilteredOutReason`), ensuring that `podEquivalenceGroups` are consistently marked before computing `ScaleUpStatus`.
```release-note
NONE

```

### DIFF
--- a/pkg/core/scaleup/orchestrator/orchestrator.go
+++ b/pkg/core/scaleup/orchestrator/orchestrator.go
@@ -1102,6 +1102,7 @@ func (o *ScaleUpOrchestrator) prepareScaleUp(ctx context.Context, args scaleUpCt
 
 	if len(options) == 0 {
 		logger.V(1).Info("No expansion options")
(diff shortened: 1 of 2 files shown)


mtrqq:
Thanks for the patch, looks good to me!

/ok-to-test
/lgtm

PR #43, opened 2026-08-10
Avoid logging negative removal latency in node latency tracker

ttetyanka:
Do not log negative removal latency when a node is removed from the unneeded list before reaching its removal threshold, as this is normal cluster operation. Also remove the verbose V(6) log for skipped metrics on non-positive latency, and ensure V(6) logging consistency.

#### What type of PR is this?
/kind cleanup

When an unneeded node becomes needed again before reaching its removal threshold (`latency <= 0`), the `NodeLatencyTracker` was logging negative latency at `klog.V(4)` (`Node "..." became needed again (unneeded for ...). Blocker: "...". Latency: -X`) and emitting a skipped-metric log at `klog.V(6)`.
Nodes becoming needed again before reaching the removal threshold is normal cluster operation (not a delayed scale-down), so emitting negative removal latency logs creates unnecessary log noise in active clusters.
This PR fixes that by:
- Guarding the `klog.V(4)` log so it only logs when `latency > 0` (when there was an actual delay beyond the threshold before becoming needed again).
- Removing the verbose `klog.V(6)` log for skipped metrics on non-positive latency.
- Ensuring `klog.V(6).Infof` is used consistently within verbosity-guarded blocks.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/core/scaledown/latencytracker/node_latency_tracker.go
+++ b/pkg/core/scaledown/latencytracker/node_latency_tracker.go
@@ -103,7 +103,7 @@ func (t *NodeLatencyTracker) Process(ctx context.Context, autoscalingCtx *ca_con
 
 	if klog.V(6).Enabled() {
 		for nodeName := range t.unneededNodes {
-			logger.Info("Node remains in unneeded list (not scaled down). Continuing to track latency.", "nodeName", nodeName)
+			logger.V(6).Info("Node remains in unneeded list (not scaled down). Continuing to track latency.", "nodeName", nodeName)
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/core/scaledown/latencytracker/node_latency_tracker.go
```diff
@@ -124,15 +124,14 @@ func (t *NodeLatencyTracker) recordAndCleanup(nodeName string, isRemoved bool) {
 
 	if latency > 0 {
 		metrics.UpdateScaleDownNodeRemovalLatency(isRemoved, delayReason, latency)
-	} else {
-		klog.V(6).Infof("Node %q was unneeded for %s (threshold %s). Latency %s is <= 0, skipping metric. isRemoved: %v, delayReason: %v",
-			nodeName, duration, info.removalThreshold, latency, isRemoved, delayReason)
-	}
-	if isRemoved {
+		if isRemoved {
```
mtrqq: Can we move logDeletion out of the condition because it happens regardless whether latency is above or below zero? While we'll still keep if !isRemoved { klog.V(4).Infof(...)) inside?

mtrqq (changes requested):
One small comment, other than that looks good to me!

pkg/core/scaledown/latencytracker/node_latency_tracker.go
ttetyanka: done

PR #44, opened 2026-08-10
Standardize UnremovableReason strings and scale-down latency metric r…

ttetyanka:
Standardize UnremovableReason strings and scale-down latency metric reasons

- Update UnremovableReason.String() to return standardized snake_case reason strings for Prometheus metrics and observability.
- Update NodeLatencyTracker to record "node_became_needed_again" when an unneeded node becomes needed again without a prior blocker.
- Update unit tests in latencytracker and metrics packages to reflect the new standardized reason strings.

#### What type of PR is this?

/kind feature
/kind cleanup

This PR changes the string representation of `simulator.UnremovableReason` to make them more user friendly .

Fixes #

* This updates the values of the `reason` label on `cluster_autoscaler_unremovable_nodes_count` and the `delay_reason` label on `cluster_autoscaler_node_removal_latency_seconds`.
* Existing dashboards or alerting rules matching on legacy PascalCase strings (e.g., `BlockedByPod`) should be updated to match the new standardized snake_case strings (e.g., `node_has_unevictable_pods`).

```release-note
Standardized `reason` and `delay_reason` Prometheus label values emitted by `cluster_autoscaler_unremovable_nodes_count` and `cluster_autoscaler_node_removal_latency_seconds` to snake_case format.
```

### DIFF
--- a/pkg/core/scaledown/latencytracker/node_latency_tracker.go
+++ b/pkg/core/scaledown/latencytracker/node_latency_tracker.go
@@ -123,7 +123,11 @@ func (t *NodeLatencyTracker) recordAndCleanup(ctx context.Context, nodeName stri
 
 	delayReason := info.latestDelayReason
 	if delayReason == "" {
-		delayReason = "none"
+		if isRemoved {
+			delayReason = "none"
+		} else {
+			delayReason = "node_became_needed_again"
+		}
 	}
 
 	if latency > 0 {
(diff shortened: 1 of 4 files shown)


mtrqq:
/assign mtrqq

While I agree that new reason representations are better - it still breaks the existing API and it should be treated as the breaking change, we are not that mature as the library yet, but I still think that we need to keep the API contracts reliable and follow kubernetes deprecation policies

How about adding another metric definition with similar name and deprecating this one and eventually removing it in couple minors as the k8s policy states: https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-metric

PR #49, opened 2026-08-12
Fix cluster-autoscaler-status update error when the size of the config map exceeds 1MB limit

damikag:
#### What type of PR is this?
/kind bug

When there are a lot of node groups, the size of the status config map grows more than 1MB. So CA fails to update the status config map and logs the following error.

```
Failed to write status configmap:
ConfigMap "cluster-autoscaler-status" is invalid: []:
Too long: may not be more than 1048576 bytes
```

This PR truncates the number of node groups from the config map to keep it under 1MB limit

When the size exceeds 1 000 000 Bytes node groups will be sorted in the ordered by state where Unhealthy > ScaleUp Backoff/Error > Active ScaleUp > Active ScaleDown. Maximum number of node groups that can be included within the 1MB limit will be find using binary search.

Alternatives Considered:
- Limit the number of node groups to a const to maintain the total size under 1MB. If we calculate an upper bound it would be around 700 node groups. This calculation includes space for the backoff field which has a error message of up to 500 bytes. This increase the variance of the upper bound and significantly lowerd the worst case node group count. 
- Increasing this 1MB limit is not possible since this comes from the size of value field in etcd.
- Altering the config map format will not solve the problem and may break some customer integrations.

Fixes #

```release-note
If the size of the cluster-autoscaler-status config map grows more than 1 000 000 Bytes list of node groups will be truncated to fit in to 1 000 000 Bytes.
```

```docs

```

### DIFF
--- a/pkg/clusterstate/utils/status.go
+++ b/pkg/clusterstate/utils/status.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
+	"sort"
 	"time"
 
 	apiv1 "k8s.io/api/core/v1"
@@ -39,6 +41,8 @@ const (
 	ConfigMapLastUpdatedKey = "cluster-autoscaler.kubernetes.io/last-updated"
 	// ConfigMapLastUpdateFormat it the timestamp format used for last update annotation in status ConfigMap
 	ConfigMapLastUpdateFormat = "2006-01-02 15:04:05.999999999 -0700 MST"
+	// maxStatusBytes is the maximum number of bytes we can write to a config map. The limit comes from 1MB value limit of etcd with a safety buffer.
+	maxStatusBytes = 1000000
 )
 
 // LogEventRecorder records events on some top-level object, to give user (without access to logs) a view of most important CA actions.
@@ -100,9 +104,9 @@ func WriteStatusConfigMap(ctx context.Context, kubeClient kube_client.Interface,
 	var errMsg string
 	maps := kubeClient.CoreV1().ConfigMaps(namespace)
 	configMap, getStatusError = maps.Get(ctx, statusConfigMapName, metav1.GetOptions{})
-	statusYaml, err := yaml.Marshal(status)
+	statusYaml, err := marshalAndTruncateStatus(status, maxStatusBytes)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to marshal status configmap: %v", err)
+		return nil, err
 	}
 	statusMsg := string(statusYaml)
 	if getStatusError == nil {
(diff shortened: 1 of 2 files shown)


pkg/clusterstate/utils/status.go
```diff
@@ -156,3 +160,101 @@ func DeleteStatusConfigMap(kubeClient kube_client.Interface, namespace string, s
 	}
 	return err
 }
+
+// marshalAndTruncateStatus marshals the ClusterAutoscalerStatus to YAML.
+// If the resulting YAML exceeds maxStatusConfigMapSize, it uses binary search
+// to find the maximum number of NodeGroups that can be included without exceeding the limit.
+func marshalAndTruncateStatus(status api.ClusterAutoscalerStatus, maxStatusConfigMapSize int) ([]byte, error) {
```
Choraden: nit: parameter maxStatusConfigMapSize shadows the global const, which might be confusing. IMO maxStatusSize would be clear enough and correspond with the function name.

pkg/clusterstate/utils/status.go
```diff
@@ -39,6 +41,8 @@ const (
 	ConfigMapLastUpdatedKey = "cluster-autoscaler.kubernetes.io/last-updated"
 	// ConfigMapLastUpdateFormat it the timestamp format used for last update annotation in status ConfigMap
 	ConfigMapLastUpdateFormat = "2006-01-02 15:04:05.999999999 -0700 MST"
+	// maxStatusConfigMapSize is the maximum number of bytes we can write to a config map. The limit comes from 1MB value limit of etcd with a safety buffer.
+	maxStatusConfigMapSize = 1000000
```
Choraden: nit: This name is not really clear to me. How about: 
maxStatusBytes (short, clear unit) or maxStatusYAMLSize (Clarifies it's for the YAML payload)

pkg/clusterstate/utils/status_test.go
```diff
@@ -259,3 +261,107 @@ func TestWriteStatusConfigMapMarshal(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			testStatus := api.ClusterAutoscalerStatus{
+				NodeGroups: make([]api.NodeGroupStatus, len(tc.nodeGroups)),
+			}
+			copy(testStatus.NodeGroups, tc.nodeGroups)
+
+			b, err := marshalAndTruncateStatus(testStatus, tc.maxSize)
+			assert.NoError(t, err)
+
+			// If the max size is extremely small (like 10), even 0 node groups will exceed it,
```
Choraden, line 366: nit: typo

pkg/clusterstate/utils/status_test.go
```diff
@@ -259,3 +261,107 @@ func TestWriteStatusConfigMapMarshal(t *testing.T) {
 	}
 	assert.YAMLEq(t, string(want), result.Data["status"])
 }
+
+func TestMarshalAndTruncateStatus(t *testing.T) {
+	testNodeGroups := []api.NodeGroupStatus{
```
Choraden, line 267: nit: could you comment rough estimates on the sizes there? This way it is easier to debug the test cases.

Choraden (commented):
This looks good to me. I left some nits, feel free to address them if you have time.

/lgtm

pkg/clusterstate/utils/status.go
damikag: Done

pkg/clusterstate/utils/status.go
damikag: Done

pkg/clusterstate/utils/status_test.go
damikag: Done

pkg/clusterstate/utils/status_test.go
damikag: Sorry where exactly?

PR #51, opened 2026-08-13
Make BenchmarkScheduleRevert run correctly

tlipski:
#### What type of PR is this?

/kind bug

`BenchmarkScheduleRevert` fails on every run today — this is not just a re-run problem.

`AddNodeInfo` pushes the claims of the pods already on a `NodeInfo` into the DRA snapshot, and
`SchedulePod` adds pods to the `NodeInfo` it schedules onto. The `NodeInfo`s are built once in
setup and shared by every `b.N` iteration and by all 120 sub-benchmarks (60 configurations × 2
snapshot stores). The second pass over them re-adds claims the first pass registered:

```
Failed to add node info: claim default/claim-... already tracked
Failed to schedule pod-0 to node-0: NodeInfo for "node-0": node not found
```

`AddNodeInfo` returns early on that error so the node is never stored, and because the failures
were reported with `b.Errorf` rather than `b.Fatalf`, each one cascaded into a hundred or so
follow-on errors.

On current `main`, a single `-benchtime=1x` pass emits 236,200 error lines (226,600 `node not
found`, 6,400 `already tracked`, 3,200 `Too many pods`) and exactly **one** of the 120
sub-benchmarks reports a timing — the first to run, before the shared state is corrupted. So the
benchmark currently measures almost nothing.

This PR:

- gives each iteration its own copies of the `NodeInfo`s, built outside the measured section
- aborts on the first failure instead of cascading
- makes the shared-claim pod names unique per node — they were all `pod-<index>`, so the
  configurations that commit instead of reverting scheduled several pods with the same key into
  one snapshot
- registers only the claims a configuration actually uses, rather than all of them
- drops the `ResetTimer` that ran before the sub-benchmarks rather than inside them

After the change the package passes in 4.8s with zero errors and all 120 sub-benchmarks report a
timing.

N/A - no issue filed for this.

Reproduce before and after with:

```
go test -run '^$' -bench BenchmarkScheduleRevert -benchtime=1x ./pkg/simulator/clustersnapshot/predicate/
```

On `main` this exits 1; with this PR it exits 0 and prints 120 results.

Two notes on choices that go slightly beyond the title:

- **`b.Errorf` → `b.Fatalf`.** A benchmark that continues after a failed setup still reports
  ns/op, so a failure surfaces as quietly wrong numbers rather than a failed run. It also fixes a
  latent nil dereference: `snapshotFactory()` returns `(nil, err)`, and the old code carried on to
  call `snapshot.SetClusterState` on the nil value.
- **Per-iteration `DeepCopy`.** This is wrapped in `StopTimer`/`StartTimer`, so the copying is not
  attributed to the measured section.

The benchmark is not part of the `ca-benchmark` CI job, which runs only
`-bench=BenchmarkRunOnce ./pkg/core/bench/...`, so nothing in CI changes as a result of this.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/simulator/clustersnapshot/predicate/predicate_snapshot_dra_benchmark_test.go
+++ b/pkg/simulator/clustersnapshot/predicate/predicate_snapshot_dra_benchmark_test.go
@@ -279,7 +279,7 @@ func BenchmarkScheduleRevert(b *testing.B) {
 		sharedClaim := createTestResourceClaim(devicesPerSlice, 1, driverName, deviceClassName)
 		sharedClaim, satisfied := allocateResourceSlicesForClaim(sharedClaim, nodeName, nodeSlice)
 		if !satisfied {
-			b.Errorf("Error during setup, claim allocation cannot be satistied")
+			b.Fatalf("Error during setup, claim allocation cannot be satisfied")
 		}
 
 		claimsOnNode := make([]*resourceapi.ResourceClaim, maxPodsCount)
@@ -297,7 +297,7 @@ func BenchmarkScheduleRevert(b *testing.B) {
 			ownedClaim = drautils.TestClaimWithPodOwnership(pod, ownedClaim)
 			ownedClaim, satisfied := allocateResourceSlicesForClaim(ownedClaim, nodeName, nodeSlice)
 			if !satisfied {
-				b.Errorf("Error during setup, claim allocation cannot be satistied")
+				b.Fatalf("Error during setup, claim allocation cannot be satisfied")
 			}
 
 			podsOnNode[podIndex] = pod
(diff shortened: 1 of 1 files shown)


pkg/simulator/clustersnapshot/predicate/predicate_snapshot_dra_benchmark_test.go
```diff
@@ -279,7 +279,7 @@ func BenchmarkScheduleRevert(b *testing.B) {
 		sharedClaim := createTestResourceClaim(devicesPerSlice, 1, driverName, deviceClassName)
 		sharedClaim, satisfied := allocateResourceSlicesForClaim(sharedClaim, nodeName, nodeSlice)
 		if !satisfied {
-			b.Errorf("Error during setup, claim allocation cannot be satistied")
+			b.Fatalf("Error during setup, claim allocation cannot be satistied")
```
Choraden: nit: there was a typo in satisfied. It'd be nice if we could fix it here :)
Similar typo one `b.Fatalf` call below.

Choraden (commented):
Thanks @tlipski! 
Please sign the CLA so that we can merge it.

/lgtm

pkg/simulator/clustersnapshot/predicate/predicate_snapshot_dra_benchmark_test.go
tlipski: no problem at all, typo fixed

Issue #37, opened 2026-08-08
DRA Provider snapshot shares ResourceClaim pointers with the informer, so a snapshot mutation writes into the informer cache


thc1006:
### What happened?

`Provider.Snapshot` builds the DRA snapshot from the ResourceClaim lister and stores those pointers directly, without copying:

```go
claims, err := p.resourceClaims.ListAll() // wraps lister.List(labels.Everything())
...
for _, claim := range claims {
    claimMap[drasnapshot.GetClaimId(claim)] = claim
}
return drasnapshot.NewSnapshot(claimMap, ...)
```

`NewSnapshot` passes the map to `NewPatchFromMap`, which keeps the same values, so the snapshot's base layer holds the same `*ResourceClaim` objects the informer lister returns. Objects from an informer lister are shared and are meant to be treated as read-only.

On a snapshot that has not been forked, the base layer is also the top layer, so `ensureClaimWritable` treats a base claim as already writable and returns it unchanged:

```go
func (s *Snapshot) ensureClaimWritable(claim *resourceapi.ResourceClaim) *resourceapi.ResourceClaim {
    if s.resourceClaims.InCurrentPatch(claimId) { // true: on an unforked snapshot the base is the top layer
        return claim // the informer object
    }
    return claim.DeepCopy()
}
```

`InCurrentPatch` checks the topmost patch, which on an unforked snapshot is the base built from the lister. `ReservePodClaims`, `UnreservePodClaims`, and the shared-claim path in `RemovePodOwnedClaims` then mutate that object in place through `AddPodReservationInPlace` and `ClearPodReservationInPlace`. The mutation lands on the informer object.

### Reproduction

A snapshot built from a single claim pointer, with `ReservePodClaims` called before any `Fork`, mutates the original object: its `Status.ReservedFor` gains the pod after the call. The shared-claim branch of `RemovePodOwnedClaims` behaves the same way. This reproduces on current main at the unit level, and fails an assertion that the source object is unchanged.

### Reachability

The ownership violation reproduces regardless of the caller. Whether a specific Cluster Autoscaler flow reaches it depends on whether that flow mutates the snapshot before forking. `addUpcomingNodesToClusterSnapshot` injects placeholder NodeInfos and `AddNodeInfo` reserves shared claims for the injected Pods, so that path looks worth checking against the fork boundaries. I have not traced every fork site, so I am reporting the ownership violation itself rather than asserting a specific production corruption.

### Expected behavior

[...]

Issue #33, opened 2026-08-07
DRA Snapshot Revert can leak claim mutations when a Pod references the same ResourceClaim more than once


thc1006:
### What happened?

`Snapshot.ensureClaimWritable` returns the caller's pointer unchanged when the claim's ID is already in the current patch, trusting that the caller passed the current value:

```go
func (s *Snapshot) ensureClaimWritable(claim *resourceapi.ResourceClaim) *resourceapi.ResourceClaim {
	claimId := GetClaimId(claim)
	if s.resourceClaims.InCurrentPatch(claimId) {
		return claim
	}
	return claim.DeepCopy()
}
```

`findPodClaims` returns the same base pointer twice when a Pod references the same ResourceClaim through two aliases, and the mutation methods call `ensureClaimWritable` once per returned entry and mutate the result in place. On the second alias the ID is already in the current patch from the first, so `ensureClaimWritable` hands back the original base pointer and the in-place mutation lands on the base object. `Revert` removes the current patch but cannot undo the base mutation, so the base layer is left dirty.

`ReservePodClaims`, `UnreservePodClaims`, and the shared-claim branch of `RemovePodOwnedClaims` all follow the same `ensureClaimWritable` then mutate-in-place then `SetCurrent` shape, so all three can leak.

### How to reproduce

Confirmed with a unit test in `pkg/simulator/dynamicresources/snapshot`. After `Fork` then `ReservePodClaims(pod)` then `Revert`, the base claim still carries the reservation:

```go
func TestReproPatchSetAliasRevertLeak(t *testing.T) {
	pod := test.BuildTestPod("aliaspod", 1, 1,
		test.WithResourceClaim("a", "shared", ""),
		test.WithResourceClaim("b", "shared", ""),
	)
	base := &resourceapi.ResourceClaim{ObjectMeta: metav1.ObjectMeta{Name: "shared", UID: "shared", Namespace: "default"}}
	snapshot := NewSnapshot(map[ResourceClaimId]*resourceapi.ResourceClaim{GetClaimId(base): base}, nil, nil, nil)

	snapshot.Fork()
	if err := snapshot.ReservePodClaims(pod); err != nil {
		t.Fatalf("ReservePodClaims: %v", err)
	}
	snapshot.Revert()

	all, _ := snapshot.ResourceClaims().List()
	for _, c := range all {
		if c.Name == "shared" && len(c.Status.ReservedFor) != 0 {
[...]

PR #55, opened 2026-08-13
DRA allocated state tracking for improved performance and memory usage

tlipski:
#### What type of PR is this?

/kind feature

The scheduler asks for the allocated device state of the cluster in every PreFilter — once per pod
placement attempt. Both `ListAllAllocatedDevices` and `GatherAllocatedState` answered that by
walking every `ResourceClaim` in the snapshot and re-deriving every `DeviceID`. Between two
consecutive attempts only a handful of claims change, so almost all of that work is repeated.

This derives the state once and then maintains it as claims are written, so the per-attempt cost
stops scaling with the size of the cluster. Every write to the claim `PatchSet` goes through
`setResourceClaim`/`deleteResourceClaim` to keep it in sync, and `Revert` re-reads only the claims
whose effective value the dropped layer changed.

This directly addresses a question @towca raised reviewing the patchset implementation in
kubernetes/autoscaler#8090:

> Have you thought about caching the LIST responses? The DRA scheduler plugin only GETs claims,
> the rest is listed (deviceclasses - which CA doesn't affect so caching would be very effective,
> resourceclaims filtered to allocated devices, resourceslices).

**Two commits, deliberately together.** The first adds `PatchSet.WalkCurrentPatchKeys`, which
exposes the keys modified or deleted in the topmost patch layer — exactly the keys whose effective
value a `Revert` changes. The second uses it. Splitting them would mean merging an exported method
with no caller, so they are in one PR.

#### Results

`benchstat`, n=6 unless noted, on a dedicated 32-vCPU cloud node (`GOMAXPROCS=8`),
using the benchmarks from #53 so both sides run byte-identical measurement code.

| benchmark | baseline | with this PR | delta |
|---|---|---|---|
| `RunOnceScaleUpDRA` | 15.003 s | 5.596 s | **-62.70%** |
| `RunOnceScaleDownDRA` (n=10) | 6.530 s | 1.938 s | **-70.32%** |
| `BinpackingEstimateDRA` 1000 nodes / 4000 claims | 526.8 ms | 100.3 ms | -80.96% |
| `BinpackingEstimateDRA` 5000 nodes / 20000 claims | 8.304 s | 1.032 s | -87.57% |
| DRA estimator geomean | 276.3 ms | 65.63 ms | -76.24% |

Memory:

| benchmark | baseline | with this PR | delta |
|---|---|---|---|
| `RunOnceScaleUpDRA` | 10.038 GiB | 3.657 GiB | -63.56% |
| `RunOnceScaleDownDRA` | 3.041 GiB | 1.308 GiB | -56.98% |

Non-DRA paths are unaffected:

| control | baseline | with this PR | delta |
|---|---|---|---|
| `RunOnceScaleUp` | 10.26 s | 10.28 s | `~` |
| `RunOnceScaleDown` | 11.84 s | 12.08 s | `~` |
| `BinpackingEstimate` (non-DRA) | 4.392 s | 4.377 s | `~` |
| snapshot store geomean | 153.8 us | 146.1 us | -4.99% |

#### How the win scales

The scenarios above move nodes, claims and pending pods together, so they cannot say which of them
the improvement tracks. Two sweeps in #53 hold one axis fixed and vary the other.

All figures below are `benchstat`, n=6, p=0.002, measured on #53 versus #53 plus this PR's two
commits — the same baseline as the tables above.

Pending pods varied, fleet fixed at 5000 nodes / 20000 allocated claims:

| pending pods | baseline | with this PR | delta |
|---|---|---|---|
| 100 | 1.711 s | 273.2 ms | -84.03% |
| 500 | 8.332 s | 1.057 s | -87.31% |
| 2500 | 43.61 s | 5.200 s | -88.08% |

Allocated claims varied, fleet fixed at 5000 nodes / 500 pending pods:

| allocated claims | baseline | with this PR | delta |
|---|---|---|---|
| 10,000 | 4.272 s | 1.035 s | -75.77% |
| 20,000 | 8.332 s | 1.057 s | -87.31% |
| 40,000 | 19.58 s | 1.137 s | -94.20% |

The second table states the mechanism directly. With this PR the cost is **flat in the number of
allocated claims** — 1.035 -> 1.057 -> 1.137 s while claims quadruple — where the baseline grows
linearly, 4.3 -> 8.3 -> 19.6 s. The saving is therefore proportional to the allocated state the old
code re-derived on every attempt, and the ratio improves as a cluster accumulates DRA workload.

#### Across device-sharing modes

`BenchmarkBinpackingEstimateDRAProfiles` (also #53) varies how devices are shaped and shared rather
than how many there are. The `noGain/*` profiles are the deliberate controls — cases constructed so
the change should *not* help.

| profile | baseline | with this PR | time | memory |
|---|---|---|---|---|
| `gpu8/exclusive` | 410.4 ms | 87.83 ms | -78.60% | -71.59% |
| `gpu8/wholeNode` | 297.6 ms | 121.7 ms | -59.12% | -48.94% |
| `gpu8/sharedClaims` | 70.67 ms | 26.14 ms | -63.00% | -52.52% |
| `mig56/fractional` | 836.6 ms | 71.78 ms | -91.42% | -88.40% |
| `devices256/multiSlice` | 840.8 ms | 64.36 ms | -92.34% | -91.99% |
| `multiDriver/gpu+net+computeDomain` | 304.1 ms | 87.34 ms | -71.27% | -67.28% |
| `noGain/existingUnallocated` | 342.7 ms | 46.49 ms | -86.43% | -86.88% |
| `noGain/manyClaimsFewPods` | 30.74 ms | 20.68 ms | -32.74% | **+6.00%** |
| `noGain/draFleetNonDraPods` | 15.96 ms | 16.37 ms | `~` | `~` |

Three things worth pulling out.

`noGain/draFleetNonDraPods` is flat on both time and memory. That is the control that matters: a
cluster with DRA capacity but pods that do not use it pays nothing for the maintained state.

`noGain/manyClaimsFewPods` is the one case that costs memory, **+6.00%** — many claims with few
placement attempts to amortise them, so the state is maintained and barely read. Time still
improves 32.74%, and the absolute figure is small, but it is the shape to watch.

`wholeNode` is the weakest real gain at -59.12%, which makes sense: one claim per node means fewer
allocated entries to re-derive per attempt, so there is less redundant work to remove.

#### What it costs

Maintaining the state is not free, and the adverse benchmarks in #53 exist to show where:

| benchmark | baseline | with this PR | delta |
|---|---|---|---|
| `AllocatedStateSingleRead/claims=100` | 69.02 us | 110.48 us | +60.05% |
| `AllocatedStateSingleRead/claims=5000` | 18.13 ms | 21.62 ms | +19.26% |
| `AllocatedStateSingleRead/claims=20000` | 96.92 ms | 111.44 ms | +14.99% |
| `AllocatedStateWriteHeavy/writes=100` | 163.7 us | 326.8 us | +99.70% |
| `AllocatedStateWriteHeavy/writes=1000` | 3.995 ms | 7.087 ms | +77.37% |
| `AllocatedStateUnallocatedClaims/claims=5000` | 2.041 ms | 1.877 ms | -8.02% |
| `AllocatedStateUnallocatedClaims/claims=20000` | 8.844 ms | 7.797 ms | -11.83% |
| `SnapshotForkRevertNoDRA` | 41.56 us | 40.27 us | `~` |

Fork/revert churn, from a dedicated high-sample run (`-benchtime=20000x -count=20`, +/-2%) because
the default sampling is too coarse at this timescale:

| benchmark | baseline | with this PR | delta | allocs/op |
|---|---|---|---|---|
| `AllocatedStateForkRevertChurn/claims=0` | 408.6 ns | 424.6 ns | +3.92% | 9 -> 9 |
| `AllocatedStateForkRevertChurn/claims=10` | 1.517 us | 2.817 us | +85.70% | 18 -> 23 |
| `AllocatedStateForkRevertChurn/claims=1000` | 1.658 us | 3.691 us | +122.62% | 18 -> 23 |

So **+2.03 us and 5 extra allocations per Fork+Revert cycle**. The overhead is *identical* at 10
claims and at 1000, because it scales with the size of the reverted patch layer, not with the
snapshot. For an empty layer it is zero — `claims=0` stays at 9 allocations, unchanged.

`PatchSet.Revert` was already O(P) in the size of the dropped layer, so the refresh keeps `Revert`
in the same complexity class with a larger constant. `BenchmarkSnapshotForkRevertNoDRA` being flat
is the check that clusters not using DRA pay nothing at all.

Net: microseconds added to fork/revert, seconds removed from the scale-up loop.

#### Cluster-scale confirmation

The benchmarks above isolate `Estimate`. To check the result survives contact with a real cluster,
the same two refs were run against a kwok-backed cluster: 1,000 nodes advertising 8 devices each,
every one of the 8,000 devices already allocated, then a 500-pod burst that forces a 63-node
scale-up.

| | baseline | with this PR | delta |
|---|---|---|---|
| `scaleUp:estimate` | 436.5 ms | 26.0 ms | **-94.0%** |
| `main` (whole CA loop) | 837.8 ms | 544.6 ms | **-35.0%** |
| time to converge | 69 s | 42 s | **-39.1%** |
| final nodes | 1,063 | 1,063 | identical |
| unschedulable pods at end | 0 | 0 | identical |

Both sides reach the same end state with nothing left pending, which is what makes the comparison
meaningful. The claims-scaling shows up here too: the same run against a 400-claim fleet gives -75%
rather than -94%.

What the scenario models, why each parameter is set the way it is, the YAML for every object it
creates, and how to rerun it:
**[hack/kwok-benchmarks/DRA-SCENARIO.md](https://github.com/tlipski/cluster-autoscaler/blob/kwok-benchmarks/hack/kwok-benchmarks/DRA-SCENARIO.md)**

This is a single run with no significance testing, so treat it as corroboration rather than primary
evidence. It does show two things the microbenchmarks structurally cannot: the whole loop moving,
and time-to-converge dropping on a real scale-up.

One caveat on reproducing it — the kwok provider cannot do DRA as it stands. `TemplateNodeInfo`
passes `nil` ResourceSlices, so a node the provider creates advertises no devices and no pod
carrying a `ResourceClaim` can ever be placed on one. The run uses a small provider fix, identical
on both sides of the comparison and not part of this PR.

#### Reproducing the comparison

The loop benchmarks already exist on `main`, so this PR can be compared against it directly:

```bash
git fetch origin main
git fetch origin pull/55/head:pr-allocated-state
go install golang.org/x/perf/cmd/benchstat@latest

git checkout origin/main
go test -run '^$' -bench BenchmarkRunOnce -benchmem -benchtime=1x -count=6 \
  ./pkg/core/bench/ > base-loop.txt

git checkout pr-allocated-state
go test -run '^$' -bench BenchmarkRunOnce -benchmem -benchtime=1x -count=6 \
  ./pkg/core/bench/ > new-loop.txt

benchstat base-loop.txt new-loop.txt
```

The DRA estimator and adverse benchmarks live in #53, so the candidate side needs both. #53 alone
is the baseline:

```bash
git fetch origin pull/53/head:pr53

git checkout -B ab-baseline pr53
git checkout -B ab-candidate pr53 && git cherry-pick main..pr-allocated-state

for side in baseline candidate; do
  git checkout ab-$side
  go test -run '^$' -bench BenchmarkBinpackingEstimateDRA -benchmem -benchtime=3x -count=6 \
    ./pkg/estimator/ > $side-dra.txt
  go test -run '^$' -bench 'BenchmarkAllocatedState|BenchmarkSnapshotForkRevertNoDRA' \
    -benchmem -benchtime=200x -count=6 \
    ./pkg/simulator/dynamicresources/snapshot/ > $side-adverse.txt
done

benchstat baseline-dra.txt candidate-dra.txt
benchstat baseline-adverse.txt candidate-adverse.txt
```

For the churn numbers specifically, raise the sampling — at 200x the readings swing by a factor of
three run to run:

```bash
go test -run '^$' -bench BenchmarkAllocatedStateForkRevertChurn -benchmem \
  -benchtime=20000x -count=20 ./pkg/simulator/dynamicresources/snapshot/
```

`RunOnceScaleDownDRA` is bimodal and does not reach significance at n=6; the -70.32% above is from
`-count=10` on that benchmark alone. All of this needs a machine with nothing else running on it —
on a laptop, thermal drift alone produced a "significant" +5% and -24% on benchmarks in code the
change does not touch.

N/A

**On withdrawal.** The tracker records what each claim contributed when it was applied, rather than
re-deriving it when the claim is withdrawn. That is not premature generality: `ensureClaimWritable`
hands back the stored object whenever the claim already lives in the current patch layer, so the
snapshot mutates claims in place and the previous allocation may already be gone by the time a
write reaches the tracker. Recording contributions keeps the derived state consistent with whatever
the claims currently say.

That in-place mutation is also the root of #33 (a Pod referencing one claim through two aliases
leaves the base layer dirty past a `Revert`) and #37 (the base layer holds the informer's own
objects). This PR neither introduces nor fixes those; the tracker is correct either way, and
`TestAllocatedStateTrackerAliasedClaim` pins down the aliasing shape from #33 specifically.

**On returned collections.** `ListAllAllocatedDevices` and `GatherAllocatedState` now return the
snapshot's own sets rather than fresh copies — that is what removes the per-attempt walk. I checked
the consumer: the DRA plugin folds them into a `structured.AllocatedState` and passes it to
`structured.NewAllocator`, which only ever reads them (`Set.Has`, range, map lookup). The contract
is documented on both methods.

**Testing.** `TestAllocatedStateTrackerMatchesReference` and
`TestAllocatedStateTrackerRandomizedOperations` check the incremental state against a from-scratch
recomputation after every operation, so the tracker cannot silently drift from what the old code
would have returned. Full package suite passes under `-race`, and the DRA benchmarks run clean
under `-race` as well.

Relates to #53, which adds the benchmarks the numbers above come from. No code dependency — this
PR can merge in either order.

```release-note
Improved Cluster Autoscaler performance for clusters using Dynamic Resource Allocation. The
allocated device state the scheduler requests on every pod placement attempt is now maintained
incrementally instead of being recomputed from every ResourceClaim, cutting DRA scale-up loop time
by roughly 60% and DRA scale-down by roughly 70% in benchmarks.
```

```docs

```

### DIFF
--- a/pkg/simulator/dynamicresources/snapshot/snapshot.go
+++ b/pkg/simulator/dynamicresources/snapshot/snapshot.go
@@ -293,10 +293,15 @@ func (s *Snapshot) Revert() {
 	// is the case for the whole life of a snapshot whenever DRA is disabled -
 	// going through allocationTracker() here would build a tracker just to ask
 	// whether it exists.
+	//
+	// IsForked is the other half of the guard. Without a Fork under it there is no layer
+	// to drop, so PatchSet.Revert is a no-op and no claim changes - but the topmost layer
+	// is then the base layer holding every claim in the snapshot, and walking it would
+	// refresh all of them to discover that nothing moved.
 	trackerBuilt := s.allocatedState != nil && s.allocatedState.isBuilt()
 
 	var revertedClaimIds []ResourceClaimId
-	if trackerBuilt {
+	if trackerBuilt && s.resourceClaims.IsForked() {
 		s.resourceClaims.WalkCurrentPatchKeys(func(claimId ResourceClaimId) bool {
 			revertedClaimIds = append(revertedClaimIds, claimId)
 			return true
--- a/pkg/simulator/dynamicresources/snapshot/snapshot.go
+++ b/pkg/simulator/dynamicresources/snapshot/snapshot.go
@@ -48,6 +48,47 @@ type Snapshot struct {
 	resourceClaims *common.PatchSet[ResourceClaimId, *resourceapi.ResourceClaim]
 	resourceSlices *common.PatchSet[string, []*resourceapi.ResourceSlice]
 	deviceClasses  *common.PatchSet[string, *resourceapi.DeviceClass]
+
+	// allocatedState is derived from resourceClaims and maintained as they change, so
+	// that the scheduler can be answered without walking every claim on every attempt.
+	// Access it through allocationTracker(), never directly - a Snapshot can be built as
+	// a struct literal (see CloneTestSnapshot) and leave this nil.
+	//
+	// Every write to resourceClaims has to go through setResourceClaim/deleteResourceClaim
+	// to keep this in sync.
+	allocatedState *allocatedStateTracker
+}
+
+// allocationTracker returns the tracker holding the state derived from the ResourceClaims,
+// creating it if needed.
+func (s *Snapshot) allocationTracker() *allocatedStateTracker {
+	if s.allocatedState == nil {
+		s.allocatedState = newAllocatedStateTracker()
+	}
+	return s.allocatedState
+}
+
+// walkResourceClaims iterates over all effective ResourceClaims in the Snapshot.
+func (s *Snapshot) walkResourceClaims(f func(*resourceapi.ResourceClaim) bool) {
+	for _, claim := range s.listResourceClaims() {
+		if !f(claim) {
+			return
+		}
+	}
+}
+
+// setResourceClaim adds or updates a ResourceClaim in the current patch layer, keeping the
+// derived allocated state in sync.
+func (s *Snapshot) setResourceClaim(claimId ResourceClaimId, claim *resourceapi.ResourceClaim) {
+	s.resourceClaims.SetCurrent(claimId, claim)
+	s.allocationTracker().apply(claimId, claim)
+}
+
+// deleteResourceClaim marks a ResourceClaim as deleted in the current patch layer, keeping
+// the derived allocated state in sync.
+func (s *Snapshot) deleteResourceClaim(claimId ResourceClaimId) {
+	s.resourceClaims.DeleteCurrent(claimId)
+	s.allocationTracker().withdraw(claimId)
 }
 
 // nonNodeLocalResourceSlicesIdentifier is a special key used in the resourceSlices patchSet
@@ -117,7 +158,7 @@ func (s *Snapshot) AddClaims(newClaims []*resourceapi.ResourceClaim) error {
 	}
 
 	for _, claim := range newClaims {
-		s.resourceClaims.SetCurrent(GetClaimId(claim), claim)
+		s.setResourceClaim(GetClaimId(claim), claim)
 	}
 
 	return nil
@@ -144,13 +185,13 @@ func (s *Snapshot) RemovePodOwnedClaims(ctx context.Context, pod *apiv1.Pod) {
 		// TODO(autoscaler/issues/9570): KEP-5729 changed IsForPod to be able to work
 		// with pod groups, re-evaluate whether they need to be considered here.
 		if err := resourceclaim.IsForPod(pod, claim, false); err == nil {
-			s.resourceClaims.DeleteCurrent(claimId)
+			s.deleteResourceClaim(claimId)
 			continue
 		}
 
 		claim := s.ensureClaimWritable(claim)
 		drautils.ClearPodReservationInPlace(claim, pod)
-		s.resourceClaims.SetCurrent(claimId, claim)
+		s.setResourceClaim(claimId, claim)
 	}
 }
 
@@ -175,7 +216,7 @@ func (s *Snapshot) ReservePodClaims(pod *apiv1.Pod) error {
 		claimId := GetClaimId(claim)
 		claim := s.ensureClaimWritable(claim)
 		drautils.AddPodReservationInPlace(claim, pod)
-		s.resourceClaims.SetCurrent(claimId, claim)
+		s.setResourceClaim(claimId, claim)
 	}
 
 	return nil
@@ -200,7 +241,7 @@ func (s *Snapshot) UnreservePodClaims(pod *apiv1.Pod) error {
 			drautils.DeallocateClaimInPlace(claim)
 		}
 
-		s.resourceClaims.SetCurrent(claimId, claim)
+		s.setResourceClaim(claimId, claim)
 	}
 	return nil
 }
@@ -233,14 +274,42 @@ func (s *Snapshot) Commit() {
 	s.deviceClasses.Commit()
 	s.resourceClaims.Commit()
 	s.resourceSlices.Commit()
+	// The effective value of every claim is unchanged by a Commit, so the derived
+	// allocated state stays valid.
 }
 
 // Revert removes the topmost patch layer for all resource types, discarding
 // any modifications or deletions recorded there.
 func (s *Snapshot) Revert() {
+	// A Revert changes the effective value of exactly the claims recorded in the layer
+	// being dropped, without any of them going through setResourceClaim. Collect them
+	// first, then re-read them once the layer is gone.
+	//
+	// This walks the dropped layer a second time, so it is O(P) in the size of that layer.
+	// PatchSet.Revert is already O(P), so the operation stays in the same complexity class
+	// and only picks up a larger constant.
+	//
+	// Skipped entirely when nothing has been derived from the claims yet, which
+	// is the case for the whole life of a snapshot whenever DRA is disabled -
+	// going through allocationTracker() here would build a tracker just to ask
+	// whether it exists.
+	trackerBuilt := s.allocatedState != nil && s.allocatedState.isBuilt()
+
+	var revertedClaimIds []ResourceClaimId
+	if trackerBuilt {
+		s.resourceClaims.WalkCurrentPatchKeys(func(claimId ResourceClaimId) bool {
+			revertedClaimIds = append(revertedClaimIds, claimId)
+			return true
+		})
+	}
+
 	s.deviceClasses.Revert()
 	s.resourceClaims.Revert()
 	s.resourceSlices.Revert()
+
+	if trackerBuilt {
+		s.allocatedState.refresh(revertedClaimIds, s.getResourceClaim)
+	}
 }
 
 // Fork adds a new, empty patch layer to the top of the stack for all
@@ -250,6 +319,7 @@ func (s *Snapshot) Fork() {
 	s.deviceClasses.Fork()
 	s.resourceClaims.Fork()
 	s.resourceSlices.Fork()
+	// An empty layer changes nothing, so the derived allocated state stays valid.
 }
 
 // listDeviceClasses retrieves all effective DeviceClasses from the snapshot.
@@ -278,7 +348,7 @@ func (s *Snapshot) listResourceClaims() []*resourceapi.ResourceClaim {
 // This is typically used internally when a claim's state (like allocation) changes.
 func (s *Snapshot) configureResourceClaim(claim *resourceapi.ResourceClaim) {
 	claimId := ResourceClaimId{Name: claim.Name, Namespace: claim.Namespace}
-	s.resourceClaims.SetCurrent(claimId, claim)
+	s.setResourceClaim(claimId, claim)
 }
 
 // getResourceClaim retrieves a specific ResourceClaim by its ID from the snapshot.
--- a/pkg/simulator/dynamicresources/snapshot/snapshot_allocated_state.go
+++ b/pkg/simulator/dynamicresources/snapshot/snapshot_allocated_state.go
@@ -18,6 +18,7 @@ package snapshot
 
 import (
 	resourceapi "k8s.io/api/resource/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/dynamic-resource-allocation/structured"
 )
@@ -66,8 +67,9 @@ type allocatedStateTracker struct {
 	// devices holds the devices claimed exclusively, matching what GatherAllocatedState
 	// reports when the DRAConsumableCapacity feature is enabled.
 	devices refCountedSet[structured.DeviceID]
-	// allDevices holds every allocated device, including the shared ones. This matches
-	// ListAllAllocatedDevices, which asks foreachAllocatedDevice to ignore sharing.
+	// allDevices holds every allocated device, including the shared ones. This is what
+	// ListAllAllocatedDevices reports, and what GatherAllocatedState reports as allocated
+	// when the DRAConsumableCapacity feature is disabled.
 	allDevices refCountedSet[structured.DeviceID]
 	// sharedDeviceIDs holds the (device, share) pairs of the shared devices.
 	sharedDeviceIDs refCountedSet[structured.SharedDeviceID]
@@ -149,7 +151,8 @@ func (t *allocatedStateTracker) apply(claimId ResourceClaimId, claim *resourceap
 			contribution.sharedDevices = append(contribution.sharedDevices, deviceID)
 			contribution.sharedDeviceIDs = append(contribution.sharedDeviceIDs, structured.MakeSharedDeviceID(deviceID, result.ShareID))
 			if result.ConsumedCapacity != nil {
-				contribution.capacity = append(contribution.capacity, structured.NewDeviceConsumedCapacity(deviceID, result.ConsumedCapacity))
+				contribution.capacity = append(contribution.capacity,
+					ownedDeviceConsumedCapacity(deviceID, result.ConsumedCapacity))
 			}
 			return
 		}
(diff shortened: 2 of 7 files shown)


tlipski:
KWOK based benchmarks did not show much evidence, digging deeper

tlipski:
kwok by default does not support DRA, so it required more work to prove but the results look sound

thc1006 (commented):
Thanks for taking this on, and for the write-up. The PreFilter re-scan is a real scaling problem, and maintaining the state incrementally through `setResourceClaim`/`deleteResourceClaim` reads as the right shape for it. The benchmark tables are also a lot more than most performance changes arrive with, so thank you for that.

I am not on the reviewer list here. I spent time in this snapshot code recently with the `findPodClaims` fix in #32, so I read through the state transitions while the assigned reviewers get to it. I traced the claim write paths, the Fork/Commit/Revert transitions, the refcount and shared-device accounting, and the feature-gate handling, and I did not find a missing write path or a current parallel Filter/Reserve race. The approach looks sound to me.

A few things stood out that I would want settled before an lgtm, left inline:

- the tracker hands back its own mutable maps and sets, which puts an ownership and lifetime expectation on `ResourceClaimTracker` that the interface does not state, and one test reads the result as a live view across mutations while the comment says it is only valid until the next change;
- the saved consumed-capacity contribution shares `resource.Quantity` storage with the claim, so an in-place update can move a value `withdraw` relies on later;
- `Revert` on an unforked snapshot walks every base claim even though `PatchSet.Revert` is a no-op at that depth;
- the branch predates the now-merged #32, and the randomized test still skips the case #32 fixed;
- the reference implementation in the test shares the new traversal helper with the code under test, so a bug in that helper would pass on both sides.

On the benchmarks, the numbers and the method are clear. The one gap I see is that they lean on #53 and #51, which are still open and not in CI, so today the result cannot be reproduced from main. Landing that chain first, or pinning the exact refs with the raw output, would make it easy to re-check later.

None of this changes the direction, which I think is a good one. I will leave the lgtm and approval to the assigned reviewers and @x13n. I can look again after a rebase if that helps.

pkg/simulator/dynamicresources/snapshot/snapshot_claim_tracker.go
```diff
@@ -44,44 +44,45 @@ func (ct snapshotClaimTracker) Get(namespace, claimName string) (*resourceapi.Re
...
+// The scheduler calls this once per PreFilter, that is once per pod scheduling attempt, so
+// the state is maintained as ResourceClaims change rather than recomputed here.
+//
+// The returned collections are the snapshot's own rather than copies: they must not be
+// modified, and they are only valid until the next change to the snapshot's ResourceClaims.
+// The DRA scheduler plugin dereferences the AllocatedState into structured.NewAllocator, and
+// the allocator only ever reads the collections (Set.Has, range, map lookup), so nothing
+// downstream writes through the alias. The AllocatedState wrapper itself is built fresh on
+// every call, so reassigning its fields below is safe.
 func (ct snapshotClaimTracker) GatherAllocatedState() (*structured.AllocatedState, error) {
```
thc1006, line 104: These results are the tracker's own maps and sets, so a caller that deletes from `AllocatedDevices` or edits `AggregatedCapacity` can pull the materialized state out of sync with the refcounts. The current DRA consumer looks read-only, and I could not find a live Filter/Reserve race, since the parallel Filter calls finish before Reserve runs. What is left is a read-only and lifetime expectation on `ResourceClaimTracker` that the interface itself does not state.

Would it make sense to put that contract at the interface boundary, or to return something the caller cannot mutate? There is also a small mismatch with `TestAllocatedStateTrackerDuplicateDevice`, which holds the returned state across two claim deletions and relies on it staying live, while the comment here says it is only valid until the next change. Settling on one meaning and making the docs, caller, and test agree would help whoever picks this up next.

pkg/simulator/dynamicresources/snapshot/snapshot_allocated_state.go
```diff
@@ -0,0 +1,273 @@
...
+	// Both device set variants are maintained, because which one a caller wants depends on
+	// the DRAConsumableCapacity feature gate, and that can be flipped after the Snapshot was
+	// created. Deriving them in one pass keeps the DeviceID interning to a single call.
+	contribution := &claimContribution{}
+	forEachAllocatedResult(claim, func(deviceID structured.DeviceID, result *resourceapi.DeviceRequestAllocationResult) {
+		if result.ShareID != nil {
+			contribution.sharedDevices = append(contribution.sharedDevices, deviceID)
+			contribution.sharedDeviceIDs = append(contribution.sharedDeviceIDs, structured.MakeSharedDeviceID(deviceID, result.ShareID))
+			if result.ConsumedCapacity != nil {
+				contribution.capacity = append(contribution.capacity, structured.NewDeviceConsumedCapacity(deviceID, result.ConsumedCapacity))
```
thc1006: This contribution is meant to be a fixed record of what the claim added, but `NewDeviceConsumedCapacity` does not deep-copy the `resource.Quantity` values. A `Quantity` can carry a shared `*inf.Dec`, so an in-place capacity update on the claim can also move the value `withdraw` reads later, and the aggregate then subtracts the new amount instead of the old one.

`DeviceConsumedCapacity.Clone()` already covers this. Would storing `NewDeviceConsumedCapacity(...).Clone()` here work? A regression with a decimal quantity like `0.1Gi`, mutated in place after it is recorded, would pin the behavior down.

pkg/simulator/dynamicresources/snapshot/snapshot.go
```diff
@@ -231,14 +272,42 @@ func (s *Snapshot) Commit() {
...
+	//
+	// Skipped entirely when nothing has been derived from the claims yet, which
+	// is the case for the whole life of a snapshot whenever DRA is disabled -
+	// going through allocationTracker() here would build a tracker just to ask
+	// whether it exists.
+	trackerBuilt := s.allocatedState != nil && s.allocatedState.isBuilt()
+
+	var revertedClaimIds []ResourceClaimId
+	if trackerBuilt {
+		s.resourceClaims.WalkCurrentPatchKeys(func(claimId ResourceClaimId) bool {
```
thc1006, line 305: `WalkCurrentPatchKeys` walks the topmost patch, and on an unforked snapshot the base is the topmost patch, while `PatchSet.Revert` is already a no-op at that depth. So `Revert` on an unforked snapshot with a built tracker refreshes every base claim even though no effective value changed, which turns an O(1) no-op into O(C*D).

Could this walk only the keys from a layer that will actually be dropped, or guard on patch depth greater than one? A small test that reverts without a `Fork` and checks that nothing is refreshed would lock it in.

pkg/simulator/dynamicresources/snapshot/snapshot_allocated_state_test.go
```diff
@@ -0,0 +1,386 @@
...
+	"k8s.io/utils/ptr"
+	drautils "sigs.k8s.io/cluster-autoscaler/pkg/simulator/dynamicresources/utils"
+
+	. "sigs.k8s.io/cluster-autoscaler/pkg/utils/test"
+)
+
+// referenceAllocatedState recomputes the allocated state from scratch, the way
+// GatherAllocatedState used to before the state was maintained incrementally. The
+// incremental tracker has to agree with this at all times.
+func referenceAllocatedState(snapshot *Snapshot, enabledConsumableCapacity bool) *structured.AllocatedState {
```
thc1006, line 54: This reads as an independent reference, but both it and the incremental tracker go through `forEachAllocatedResult`, which this PR introduces. A bug in the result traversal, `DeviceID` construction, or shared-result classification would then show up on both sides and the comparison would still pass.

Would it be worth deriving the expected state here straight from `claim.Status.Allocation.Devices.Results`, independent of the production helper? A few table-driven expected states for the dedicated, shared, duplicate-share, and consumed-capacity cases would also catch things one random walk can miss.

pkg/simulator/dynamicresources/snapshot/snapshot_allocated_state_test.go
```diff
@@ -0,0 +1,386 @@
...
+					}
+				case 3: // unreserve - can deallocate the claim in place
+					if err := snapshot.UnreservePodClaims(pod); err != nil {
+						break
+					}
+				case 4: // remove the pod-owned claim
+					if _, found := snapshot.getResourceClaim(claimId); !found {
+						// findPodClaims returns nil entries for claims that are no longer
+						// tracked, which RemovePodOwnedClaims dereferences. That is fixed
+						// separately - skip the case here rather than depend on it.
```
thc1006: #32 landed on main last week, so this workaround is now stale. On a rebase, dropping the skip and letting the randomized sequence reach the missing-claim path would exercise exactly the transition #32 fixed, which is worth keeping covered. The current green CI run also predates that merge, so a rerun on the rebased base would be worth it.

pkg/simulator/dynamicresources/snapshot/snapshot_claim_tracker.go
tlipski: Yes, that's a great point. I have updated the comments about this and also attempted to track the changes with a tripwire. Test is updated as well - no doubt here.

The goal of returning raw data is avoiding unnecessary copies for something that will be just read.

Since the interface uses generic map, a proper fix (returning immutable data structure type) would require the change in contract for cluster-autoscaler, which I feel like goes over the scope of this PR. If there is a neat way to address this that I have missed, please point me to it.

_Please note that there is already (bad?) precedent for returning (and modifying) internal data in the codebase - where DeltaSnapshotStore hands out a memoized internal slice through a kube-scheduler lister interface, for the same reason and with the same unstated contract. Definitely outside of the scope for this PR though._

https://github.com/kubernetes-sigs/cluster-autoscaler/blob/8726c7a4437f5a7dd56bfd6c77710adfefa47fcd/pkg/simulator/clustersnapshot/store/delta.go#L337-L339
https://github.com/kubernetes-sigs/cluster-autoscaler/blob/8726c7a4437f5a7dd56bfd6c77710adfefa47fcd/pkg/simulator/clustersnapshot/store/delta.go#L172-L174

The memoize techniques also fall into this pattern, e.g.

https://github.com/kubernetes-sigs/cluster-autoscaler/blob/8726c7a4437f5a7dd56bfd6c77710adfefa47fcd/pkg/simulator/clustersnapshot/store/delta.go#L342-L357

pkg/simulator/dynamicresources/snapshot/snapshot_allocated_state.go
tlipski: Addressed for both, although Clone()'s cost +12.95% time / +23.14% allocs on the write path, because the constructor builds a map of pointers and Clone immediately discards it for a second. Building the owned copy in one pass gets that to +1.57% / +5.01%, with the int64 path free.

Again, probably it would be best to fix in NewDeviceConsumedCapacity(), which again requires a change of contract. So far added a solution in the snapshot_allocated_state.go with appropriate comments.

pkg/simulator/dynamicresources/snapshot/snapshot.go
tlipski: Great catch, fixed.

pkg/simulator/dynamicresources/snapshot/snapshot_allocated_state_test.go
tlipski: Yes, fixed. It also allowed to remove the unused `foreachAllocatedDevice`.

pkg/simulator/dynamicresources/snapshot/snapshot_allocated_state_test.go
tlipski: rebased, thank you!

tlipski:
@thc1006 thank you so much for the detailed review, these are great points. I have pushed the changes that address them and responded in the comments. Feel free to resolve the conversations where there is no concern anymore

pkg/simulator/dynamicresources/snapshot/snapshot_claim_tracker.go
thc1006: No, nothing neater that fits this PR. The two conditions you wrote down are the contract, and the allocator test pins the one caller that exists. The durable fix is where you point, in schedulerapi, and that's its own change. Thanks for writing down what the guard does not catch as plainly as what it does; that part usually gets left out. I checked the rest on 3151d38 as well, all good.

pkg/simulator/dynamicresources/snapshot/snapshot_allocated_state.go
thc1006: The one-pass copy is a better answer than the Clone I suggested; +1.57% on the write path against +12.95% settles it. Agreed the constructor is where it belongs eventually.

PR #63, opened 2026-08-17
Prevent atomic node groups from being unnecessarily categorized as unneeded

pwitkiewicz:
#### What type of PR is this?

/kind bug
/kind feature

This PR refactors scale-down simulation for Atomic Node Pools (ZeroOrMaxNodeScaling=true) to evaluate candidate groups as all-or-nothing units with early-abort while preserving fair interleaving between atomic and non-atomic node groups.

Previously, the scale-down planner simulated atomic node pools node-by-node in arbitrary order:
1. If a node late in an atomic node group failed removal simulation (e.g., due to an unevictable pod or PDB limit), all earlier simulations for that group were wasted.
2. Nodes from partially removable atomic groups were added to removableList (unneededNodes) and registered in NodeLatencyTracker. When later in the scaled down process node group was rejected from scaling down because of being incomplete, those partially simulated nodes remained trapped in tracking across cycles, ticking past the 600s threshold and triggering false SLO latency alerts.

Fixes b/524606761

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/pkg/core/scaledown/planner/planner.go
+++ b/pkg/core/scaledown/planner/planner.go
@@ -41,6 +41,7 @@ import (
 	"sigs.k8s.io/cluster-autoscaler/pkg/simulator/options"
 	"sigs.k8s.io/cluster-autoscaler/pkg/simulator/scheduling"
 	"sigs.k8s.io/cluster-autoscaler/pkg/simulator/utilization"
+	"sigs.k8s.io/cluster-autoscaler/pkg/utils/atomic"
 	"sigs.k8s.io/cluster-autoscaler/pkg/utils/errors"
 	pod_util "sigs.k8s.io/cluster-autoscaler/pkg/utils/pod"
 )
@@ -294,28 +295,44 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
 	unremovableCount := 0
 	var removableList []simulator.NodeToBeRemoved
 	atomicScaleDownNodesCount := 0
+	nonAtomicRemovableCount := 0
 	p.unremovableNodes.Update(ctx, p.autoscalingCtx.ClusterSnapshot, p.latestUpdate)
 	currentlyUnneededNodeNames, utilizationMap, ineligible := p.eligibilityChecker.FilterOutUnremovable(ctx, p.autoscalingCtx, scaleDownCandidates, p.latestUpdate, p.unremovableNodes)
 	for _, n := range ineligible {
 		p.unremovableNodes.Add(n)
 	}
 	p.nodeUtilizationMap = utilizationMap
+
+	// Prefilter incomplete atomic node groups so simulation isn't wasted on atomic groups that cannot scale down.
+	var prefilteredUnremovableCount int
+	currentlyUnneededNodeNames, prefilteredUnremovableCount = p.prefilterIncompleteAtomicNodeGroups(ctx, currentlyUnneededNodeNames, unremovableTimeout)
+	unremovableCount += prefilteredUnremovableCount
+
 	timer := time.NewTimer(p.autoscalingCtx.ScaleDownSimulationTimeout)
 	var skippedNodes []string
 
-	for i, node := range currentlyUnneededNodeNames {
+	for i, nodeName := range currentlyUnneededNodeNames {
 		if timedOut(timer) {
-			skippedNodes = currentlyUnneededNodeNames[i:]
-			logger.Info("Some nodes skipped in scale down simulation due to timeout", "skippedNodesCount", len(currentlyUnneededNodeNames)-i, "nodesCount", len(currentlyUnneededNodeNames))
-			break
-		}
-		if len(removableList)-atomicScaleDownNodesCount >= p.unneededNodesLimit() {
-			skippedNodes = currentlyUnneededNodeNames[i:]
-			logger.V(4).Info("Some nodes skipped in scale down simulation: unneeded nodes count already exceeded limit so no point in looking for more.", "skippedNodesCount", len(currentlyUnneededNodeNames)-i, "nodesCount", len(currentlyUnneededNodeNames), "unneededNodesCount", len(removableList), "atomicScaleDownNodesCount", atomicScaleDownNodesCount)
+			skippedNodes = append(skippedNodes, currentlyUnneededNodeNames[i:]...)
+			logger.Info("Some nodes skipped in scale down simulation due to timeout.", "skippedNodesCount", len(currentlyUnneededNodeNames)-i, "nodesCount", len(currentlyUnneededNodeNames))
 			break
 		}
 
-		removable, unremovable := p.rs.SimulateNodeRemoval(ctx, node, podDestinations, p.latestUpdate, p.autoscalingCtx.RemainingPdbTracker)
+		nodeInfo, err := p.autoscalingCtx.ClusterSnapshot.GetNodeInfo(nodeName)
+		if err != nil || nodeInfo == nil || nodeInfo.Node() == nil {
+			logger.Error(err, "Failed to get node info", "nodeName", nodeName)
+			continue
+		}
+		node := nodeInfo.Node()
+		_, isAtomic := atomic.IsAtomicNodeGroup(ctx, p.autoscalingCtx, node)
+
+		if !isAtomic && nonAtomicRemovableCount >= p.unneededNodesLimit() {
+			skippedNodes = append(skippedNodes, nodeName)
+			logger.V(4).Info("Skipping non-atomic node in scale down simulation: there are already some non-atomic unneeded nodes.", "node", klog.KObj(node), "nonAtomicUnneededNodesCount", nonAtomicRemovableCount)
+			continue
+		}
+
+		removable, unremovable := p.rs.SimulateNodeRemoval(ctx, nodeName, podDestinations, p.latestUpdate, p.autoscalingCtx.RemainingPdbTracker)
 		if removable != nil {
 			_, inParallel, _ := p.autoscalingCtx.RemainingPdbTracker.CanRemovePods(removable.PodsToReschedule)
 			if !inParallel {
@@ -324,9 +341,11 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
 			delete(podDestinations, removable.Node.Name)
 			p.autoscalingCtx.RemainingPdbTracker.RemovePods(removable.PodsToReschedule)
 			removableList = append(removableList, *removable)
-			if p.atomicScaleDownNode(ctx, removable) {
+			if isAtomic {
 				atomicScaleDownNodesCount++
-				logger.V(2).Info("Considering node for atomic scale down", "node", klog.KObj(removable.Node), "atomicScaleDownNodesCount", atomicScaleDownNodesCount)
+				logger.V(2).Info("Considering node for atomic scale down.", "node", klog.KObj(node), "currentAtomicScaleDownNodesCount", atomicScaleDownNodesCount)
+			} else {
+				nonAtomicRemovableCount++
 			}
 		}
 		if unremovable != nil {
@@ -334,34 +353,97 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
 			p.unremovableNodes.AddTimeout(unremovable, unremovableTimeout)
 		}
 	}
+
 	p.handleUnprocessedNodes(skippedNodes)
 	p.unneededNodes.Update(ctx, p.autoscalingCtx, removableList, p.latestUpdate)
 	if unremovableCount > 0 {
-		logger.V(1).Info("Some nodes found to be unremovable in simulation, will re-check them within timeout", "unremovableNodesCount", unremovableCount, "timeout", unremovableTimeout)
+		logger.V(1).Info("Some nodes found to be unremovable in simulation, will re-check them within timeout.", "unremovableNodesCount", unremovableCount, "timeout", unremovableTimeout)
 	}
 }
 
-// atomicScaleDownNode checks if the removable node would be considered for atomic scale down.
-func (p *Planner) atomicScaleDownNode(ctx context.Context, node *simulator.NodeToBeRemoved) bool {
+// prefilterIncompleteAtomicNodeGroups identifies atomic node groups in currentlyUnneededNodeNames.
+// If an atomic node group has fewer unneeded candidates than its target size, it cannot be
+// atomically scaled down. Such incomplete atomic node groups are filtered out of simulation,
+// and their nodes are marked as unremovable.
+func (p *Planner) prefilterIncompleteAtomicNodeGroups(ctx context.Context, nodeNames []string, unremovableTimeout time.Time) ([]string, int) {
 	logger := klog.FromContext(ctx)
-	nodeGroup, err := p.autoscalingCtx.CloudProvider.NodeGroupForNode(ctx, node.Node)
+	atomicGroupNodes := make(map[string][]string)
+	atomicGroupObj := make(map[string]cloudprovider.NodeGroup)
+	nodeNameToNode := make(map[string]*apiv1.Node)
+
+	for _, nodeName := range nodeNames {
+		nodeInfo, err := p.autoscalingCtx.ClusterSnapshot.GetNodeInfo(nodeName)
+		if err != nil || nodeInfo == nil || nodeInfo.Node() == nil {
+			continue
+		}
+		node := nodeInfo.Node()
+		nodeNameToNode[nodeName] = node
+		nodeGroup, isAtomic := atomic.IsAtomicNodeGroup(ctx, p.autoscalingCtx, node)
+		if isAtomic && nodeGroup != nil {
+			ngID := nodeGroup.Id()
+			atomicGroupNodes[ngID] = append(atomicGroupNodes[ngID], nodeName)
+			atomicGroupObj[ngID] = nodeGroup
+		}
+	}
+
+	if len(atomicGroupNodes) == 0 {
+		return nodeNames, 0
+	}
+
+	incompleteNodes := make(map[string]bool)
+	unremovableCount := 0
+
+	allNodes, err := allNodes(p.autoscalingCtx.ClusterSnapshot)
 	if err != nil {
-		logger.Error(err, "Failed to get node info", "node", klog.KObj(node.Node))
-		return false
+		return nil, 0
 	}
-	if nodeGroup == nil {
-		logger.Error(nil, "Node group not found for node", "node", klog.KObj(node.Node))
-		return false
+
+	for ngID, unneededNodes := range atomicGroupNodes {
+		ng := atomicGroupObj[ngID]
+		targetSize, err := ng.TargetSize(ctx)
+		if err != nil {
+			logger.Error(err, "Failed to get target size for node group", "nodeGroupId", ng.Id())
+			continue
+		}
+
+		if len(unneededNodes) < targetSize {
+			registeredCount, countErr := atomic.CountRegisteredNodesForGroup(ctx, ng, allNodes)
+			if countErr == nil && len(unneededNodes) >= registeredCount {
+				// All registered nodes in the snapshot are unneeded
+				continue
+			}
+			logger.V(2).Info("Atomic node group has some unneeded nodes. Filtering out from scale down simulation.", "nodeGroupId", ng.Id(), "ngUnneededNodesCount", len(unneededNodes), "ngTargetSize", targetSize)
+			for _, nodeName := range unneededNodes {
+				incompleteNodes[nodeName] = true
+				if node, ok := nodeNameToNode[nodeName]; ok {
+					p.unremovableNodes.AddTimeout(&simulator.UnremovableNode{
+						Node:   node,
+						Reason: simulator.AtomicScaleDownFailed,
+					}, unremovableTimeout)
+					unremovableCount++
+				}
+			}
+		}
 	}
-	autoscalingOptions, err := nodeGroup.GetOptions(ctx, p.autoscalingCtx.NodeGroupDefaults)
-	if err != nil && err != cloudprovider.ErrNotImplemented {
-		logger.Error(err, "Failed to get autoscaling options for node group", "nodeGroupId", nodeGroup.Id())
-		return false
+
+	if len(incompleteNodes) == 0 {
+		return nodeNames, 0
 	}
-	if autoscalingOptions != nil && autoscalingOptions.ZeroOrMaxNodeScaling {
-		return true
+
+	filteredNodeNames := make([]string, 0, len(nodeNames)-len(incompleteNodes))
+	for _, nodeName := range nodeNames {
+		if !incompleteNodes[nodeName] {
+			filteredNodeNames = append(filteredNodeNames, nodeName)
+		}
 	}
-	return false
+
+	return filteredNodeNames, unremovableCount
+}
+
+// isNodeAtomicScaleDown checks if the node would be considered for atomic scale down.
+func (p *Planner) isNodeAtomicScaleDown(ctx context.Context, node *apiv1.Node) bool {
+	_, isAtomic := atomic.IsAtomicNodeGroup(ctx, p.autoscalingCtx, node)
+	return isAtomic
 }
 
 // unneededNodesLimit returns the number of nodes after which calculating more
--- a/pkg/core/scaledown/planner/planner.go
+++ b/pkg/core/scaledown/planner/planner.go
@@ -24,7 +24,6 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	klog "k8s.io/klog/v2"
-	"sigs.k8s.io/cluster-autoscaler/pkg/cloudprovider"
 	ca_context "sigs.k8s.io/cluster-autoscaler/pkg/context"
 	"sigs.k8s.io/cluster-autoscaler/pkg/core/scaledown"
 	"sigs.k8s.io/cluster-autoscaler/pkg/core/scaledown/eligibility"
@@ -53,6 +52,7 @@ type eligibilityChecker interface {
 type removalSimulator interface {
 	DropOldHints()
 	SimulateNodeRemoval(ctx context.Context, node string, podDestinations map[string]bool, timestamp time.Time, remainingPdbTracker pdb.RemainingPdbTracker) (*simulator.NodeToBeRemoved, *simulator.UnremovableNode)
+	SimulateNodesGroupRemoval(ctx context.Context, nodeNames []string, destinationMap map[string]bool, timestamp time.Time, remainingPdbTracker pdb.RemainingPdbTracker) ([]simulator.NodeToBeRemoved, *simulator.UnremovableNode, int)
 }
 
 // controllerReplicasCalculator calculates a number of target and expected replicas for a given controller.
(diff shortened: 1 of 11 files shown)


pkg/core/scaledown/planner/planner.go
```diff
@@ -294,28 +295,67 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
+
+		if isAtomic && nodeGroup != nil {
+			ngID := nodeGroup.Id()
+			if processedAtomicGroups[ngID] {
+				continue
+			}
+			groupRemovable, groupUnremovableCount, groupSkipped, timedOut := p.simulateAtomicNodeGroup(ctx, ngID, atomicGroups[ngID], podDestinations, unremovableTimeout, timer)
+			if timedOut {
+				skippedNodes = p.appendUnprocessed(ctx, skippedNodes, currentlyUnneededNodeNames[i:], processedAtomicGroups)
+				klog.Warningf("%d out of %d nodes skipped in scale down simulation due to timeout.", len(skippedNodes), len(currentlyUnneededNodeNames))
```
norbertcyran: we migrated recently to contextual logging, could you please adjust the logs? https://github.com/kubernetes-sigs/cluster-autoscaler/blob/main/pkg/proposals/structural-contextual-logging.md

pkg/core/scaledown/planner/planner.go
```diff
@@ -324,44 +364,224 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
 			delete(podDestinations, removable.Node.Name)
 			p.autoscalingCtx.RemainingPdbTracker.RemovePods(removable.PodsToReschedule)
 			removableList = append(removableList, *removable)
```
norbertcyran: this entire branch is now skipped in atomic node groups. Is that expected? If yes, why?

pkg/core/scaledown/planner/planner.go
```diff
@@ -294,28 +295,67 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
+				removableList = append(removableList, groupRemovable...)
+				atomicScaleDownNodesCount += len(groupRemovable)
+			}
+			unremovableCount += groupUnremovableCount
+			skippedNodes = append(skippedNodes, groupSkipped...)
+			continue
 		}
 
-		removable, unremovable := p.rs.SimulateNodeRemoval(ctx, node, podDestinations, p.latestUpdate, p.autoscalingCtx.RemainingPdbTracker)
+		if nonAtomicRemovableCount >= p.unneededNodesLimit() {
```
norbertcyran, line 351: will atomic node groups no more count towards the limit?

pkg/core/scaledown/planner/planner.go
```diff
@@ -324,44 +364,224 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
+			logger.Error(err, "Failed to get target size for node group", "nodeGroupId", ng.Id())
+			continue
+		}
+
+		if len(unneededNodes) < targetSize {
+			registeredCount, countErr := atomic.CountRegisteredNodesForGroup(ctx, ng, allNodes)
+			if countErr == nil && len(unneededNodes) >= registeredCount {
+				// All registered nodes in the snapshot are unneeded
+				continue
+			}
```
norbertcyran: I'm not sure if I fully understand this part. If all registered nodes are unneeded, but unneededNodes is less that targetSize, wouldn't that mean that there are upcoming nodes? If there are upcoming nodes, we should not continue with the atomic scale up

Also, if you take a look at the call hierarchy, `nodeNames` is populated from the node lister, so it will contain only registered nodes. I might be missing something, but it seems to me that everything other than `if len(unneededNodes) < targetSize` is redundant

pkg/core/scaledown/planner/planner.go
```diff
@@ -294,28 +295,67 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
 	p.unremovableNodes.Update(ctx, p.autoscalingCtx.ClusterSnapshot, p.latestUpdate)
 	currentlyUnneededNodeNames, utilizationMap, ineligible := p.eligibilityChecker.FilterOutUnremovable(ctx, p.autoscalingCtx, scaleDownCandidates, p.latestUpdate, p.unremovableNodes)
 	for _, n := range ineligible {
 		p.unremovableNodes.Add(n)
 	}
 	p.nodeUtilizationMap = utilizationMap
+
+	// Prefilter incomplete atomic node groups so simulation isn't wasted on atomic groups that cannot scale down.
+	var prefilteredUnremovableCount int
+	currentlyUnneededNodeNames, prefilteredUnremovableCount = p.prefilterIncompleteAtomicNodeGroups(ctx, currentlyUnneededNodeNames, unremovableTimeout)
```
norbertcyran: nit: wouldn't it be cleaner to move this logic to `FilterOutUnremovable`?

pkg/core/scaledown/planner/planner.go
```diff
@@ -324,44 +364,224 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
+		skippedNodes = append(skippedNodes, nodeName)
+	}
+	return skippedNodes
+}
+
+// simulateAtomicNodeGroup simulates removal of all nodes in an atomic node group as an all-or-nothing unit.
+// If any node in the group is unremovable, simulation of remaining nodes is aborted early, and any temporary
+// podDestinations and RemainingPdbTracker mutations are rolled back.
+// Returns removableList for the group, unremovableCount, skippedNodes, and whether simulation timed out.
+func (p *Planner) simulateAtomicNodeGroup(
```
norbertcyran, line 401: nit: how about moving it (or parts of it) to RemovalSimulator?

pkg/core/scaledown/planner/planner.go
```diff
@@ -294,28 +295,67 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
+	atomicGroups := p.groupAtomicNodes(ctx, currentlyUnneededNodeNames)
+	processedAtomicGroups := make(map[string]bool)
 
-	for i, node := range currentlyUnneededNodeNames {
+	for i, nodeName := range currentlyUnneededNodeNames {
 		if timedOut(timer) {
-			skippedNodes = currentlyUnneededNodeNames[i:]
-			logger.Info("Some nodes skipped in scale down simulation due to timeout", "skippedNodesCount", len(currentlyUnneededNodeNames)-i, "nodesCount", len(currentlyUnneededNodeNames))
+			skippedNodes = p.appendUnprocessed(ctx, skippedNodes, currentlyUnneededNodeNames[i:], processedAtomicGroups)
+			logger.Info("Some nodes skipped in scale down simulation due to timeout.", "skippedNodesCount", len(skippedNodes), "nodesCount", len(currentlyUnneededNodeNames))
```
norbertcyran: as we do the same thing here and in L338-339. can we move it out of the for loop to avoid duplication?

pkg/core/scaledown/planner/planner.go
```diff
@@ -294,28 +295,67 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
 		p.unremovableNodes.Add(n)
 	}
 	p.nodeUtilizationMap = utilizationMap
+
+	// Prefilter incomplete atomic node groups so simulation isn't wasted on atomic groups that cannot scale down.
+	var prefilteredUnremovableCount int
+	currentlyUnneededNodeNames, prefilteredUnremovableCount = p.prefilterIncompleteAtomicNodeGroups(ctx, currentlyUnneededNodeNames, unremovableTimeout)
+	unremovableCount += prefilteredUnremovableCount
+
 	timer := time.NewTimer(p.autoscalingCtx.ScaleDownSimulationTimeout)
```
norbertcyran: instead of using timer, we should consider using a context with timeout. As simulations can be long-running (AFAIK 30s by default), they can interfere with graceful shutdown. If we checked on ctx.Done() instead of the timer, simulation could be cancelled also when CA receives a termination signal. Also `simulateAtomicNodeGroup` could return a standard error instead of `timedOut` bool. WDYT?

pkg/core/scaledown/planner/planner.go
```diff
@@ -324,44 +364,224 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
+		if removable != nil {
+			_, inParallel, _ := p.autoscalingCtx.RemainingPdbTracker.CanRemovePods(removable.PodsToReschedule)
+			if !inParallel {
+				removable.IsRisky = true
+			}
+			delete(podDestinations, removable.Node.Name)
+			p.autoscalingCtx.RemainingPdbTracker.RemovePods(removable.PodsToReschedule)
+			groupRemovable = append(groupRemovable, *removable)
+			continue
+		}
```
norbertcyran: this might be problematic (probably always has been). On one hand, we call `CanRemovePods` just to check if a removable is risky. However, [SimulateNodeRemoval](https://github.com/kubernetes-sigs/cluster-autoscaler/blob/4110ef5a1fb7c02838764a173687b16130dc5e38/pkg/simulator/cluster.go#L198) calls [GetPodsToMove](https://github.com/kubernetes-sigs/cluster-autoscaler/blob/4110ef5a1fb7c02838764a173687b16130dc5e38/pkg/simulator/drain.go#L50), which in turn checks drainability rules, including [pdb rule](https://github.com/kubernetes-sigs/cluster-autoscaler/blob/main/pkg/simulator/drainability/rules/pdb/rule.go). This is something to be confirmed, but I'm afraid that this can block atomic scale down if draining all the nodes at once would exceed PDB

pkg/core/scaledown/planner/planner.go
```diff
@@ -324,44 +364,224 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
+			filteredNodeNames = append(filteredNodeNames, nodeName)
+		}
 	}
-	return false
+
+	return filteredNodeNames, unremovableCount
+}
+
+// groupAtomicNodes groups unneeded atomic node names by their NodeGroup ID.
+func (p *Planner) groupAtomicNodes(ctx context.Context, nodeNames []string) map[string][]string {
```
norbertcyran: nit: let's move it to utils/atomic

norbertcyran (commented):
I will finish the review tomorrow, as the PR is a bit too large for a single round. Adding some comments so you can address them in the meantime

pkg/core/scaledown/planner/planner.go
```diff
@@ -324,44 +364,224 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
+			continue
+		}
+
+		if len(unneededNodes) < targetSize {
+			registeredCount, countErr := atomic.CountRegisteredNodesForGroup(ctx, ng, allNodes)
+			if countErr == nil && len(unneededNodes) >= registeredCount {
+				// All registered nodes in the snapshot are unneeded
+				continue
+			}
+			logger.V(2).Info("Atomic node group has some unneeded nodes. Filtering out from scale down simulation.", "nodeGroupId", ng.Id(), "ngUnneededNodesCount", len(unneededNodes), "ngTargetSize", targetSize)
```
norbertcyran: nit: I think this would be clearer if it said "Atomic node group has some unremovable nodes"

pkg/core/scaledown/planner/planner.go
pwitkiewicz: missed that one, will fix

pkg/core/scaledown/planner/planner.go
pwitkiewicz: hey, IIUC atomic nodes never counted towards the limit, in original `planner.go` there is line:

`if len(removableList)-atomicScaleDownNodesCount >= p.unneededNodesLimit()`

pkg/core/scaledown/planner/planner.go
```diff
@@ -324,44 +364,224 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
+			for _, nodeName := range unneededNodes {
+				incompleteNodes[nodeName] = true
+				if node, ok := nodeNameToNode[nodeName]; ok {
+					p.unremovableNodes.AddTimeout(&simulator.UnremovableNode{
+						Node:   node,
+						Reason: simulator.AtomicScaleDownFailed,
+					}, unremovableTimeout)
```
norbertcyran: I wonder if atomic node groups shouldn't have a little longer timeout than non-atomic nodes. WDYT?

pkg/core/scaledown/planner/planner.go
pwitkiewicz: good point, extracted it to a closure

pkg/core/scaledown/planner/planner.go
pwitkiewicz: > If there are upcoming nodes, we should not continue with the atomic scale up

I assume you meant scale down, not scale up. It's a fair question but there is an edge case we should handle - broken upcoming nodes. With accelerator slices it might happen that if a node group attempted to scale up to targetSize (e.g. 10), but 2 VMs failed to provision / failed to register with Kubernetes and are permanently broken or stuck, targetSize remains 10 while registeredCount is only 8.

> but it seems to me that everything other than if len(unneededNodes) < targetSize is redundant

everything you mean what exactly? I've changed `len(unneededNodes) >= registeredCount` to `len(unneededNodes) == registeredCount`, 

if all registered nodes in the group are unneeded, we should allow it through simulation so that `scale_down_set_processor` can eventually clean up and scale down the broken group rather than leaving it stuck forever

pkg/core/scaledown/planner/planner.go
pwitkiewicz: done

pkg/core/scaledown/planner/planner.go
pwitkiewicz: refactored

pkg/core/scaledown/planner/planner.go
pwitkiewicz: I guess we can take care of it in a way that if a PDB is already at 0 before starting simulation, the pod genuinely cannot be evicted right now. But if budget was >0 initially and was only exhausted by a sibling node in the same atomic group, it should mark the removal as IsRisky (sequential drain needed) rather than hard-failing the atomic group. WDYT?

pkg/core/scaledown/planner/planner.go
pwitkiewicz: done

pkg/core/scaledown/planner/planner.go
pwitkiewicz: really good point! thanks, I will try to do that

pkg/simulator/cluster.go
```diff
@@ -220,6 +221,69 @@ func (r *RemovalSimulator) SimulateNodeRemoval(
...
+		if remainingPdbTracker != nil {
+			if err := remainingPdbTracker.SetPdbs(pdbsSnapshot); err != nil {
+				logger.Error(err, "Failed to restore PDB snapshot after rollback in group removal simulation")
+			}
+		}
+	}
+
+	var removableList []NodeToBeRemoved
+	for j, nodeName := range nodeNames {
+		removable, unremovable := r.SimulateNodeRemoval(ctx, nodeName, destinationMap, timestamp, remainingPdbTracker)
```
norbertcyran, line 266: shouldn't we check `ctx.Err()` in the beginning of the loop? IIRC, in the previous revision simulation was cancelled when the timeout was reached

pkg/core/scaledown/planner/planner.go
```diff
@@ -324,44 +363,81 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
+// If any node in the group is unremovable, simulation of remaining nodes is aborted early, and any temporary
+// podDestinations and RemainingPdbTracker mutations are rolled back.
+// Returns removableList for the group, unremovableCount, skippedNodes, and an error if the context was canceled or timed out.
+func (p *Planner) simulateAtomicNodeGroup(
+	ctx context.Context,
+	ngID string,
+	ngNodes []string,
+	podDestinations map[string]bool,
+	unremovableTimeout time.Time,
+) ([]simulator.NodeToBeRemoved, int, []string, error) {
```
norbertcyran, line 407: nit: use named return parameters or wrap `[]simulator.NodeToBeRemoved, int, []string` in a struct. I know the comment clarifies what the return parameters are, but ideally they should be self-descriptive

pkg/core/scaledown/planner/planner.go
```diff
@@ -324,44 +363,81 @@ func (p *Planner) categorizeNodes(ctx context.Context, podDestinations map[strin
...
-	if autoscalingOptions != nil && autoscalingOptions.ZeroOrMaxNodeScaling {
-		return true
+
+	groupRemovable, unremovable, failedIdx := p.rs.SimulateNodesGroupRemoval(ctx, ngNodes, podDestinations, p.latestUpdate, p.autoscalingCtx.RemainingPdbTracker)
+	if unremovable != nil {
+		logger.V(2).Info("Atomic node group cannot scale down because node is unremovable. Early aborting simulation for nodes.", "nodeGroupId", ngID, "nodeName", ngNodes[failedIdx], "ngNodesCount", len(ngNodes))
+		for idx, name := range ngNodes {
+			if idx == failedIdx {
+				p.unremovableNodes.AddTimeout(unremovable, unremovableTimeout)
+			} else {
```
norbertcyran, line 420: nit: do we need `failedIdx`? We can just check `if name == unremovable.Node.Name`

pkg/core/scaledown/planner/planner.go
norbertcyran: >I assume you meant scale down, not scale up

yes, that's right

>2 VMs failed to provision / failed to register with Kubernetes and are permanently broken or stuck, targetSize remains 10 while registeredCount is only 8.

CA is responsible for cleaning up such nodes: https://github.com/kubernetes-sigs/cluster-autoscaler/blob/25cf3ae57e85ef1d2710b561f05821f06dee8bc1/pkg/core/static_autoscaler.go#L481. That should be eventually consistent, the actual size shouldn't be different from the target size for a prolonged period. There's another function ensuring that: https://github.com/kubernetes-sigs/cluster-autoscaler/blob/64f225cac68f28c6357d0c9fe6a303751de9d54b/pkg/core/static_autoscaler.go#L980

>everything you mean what exactly?

I mean that check `if len(unnededNodes) < targetSize` should be sufficient. I don't believe there's a need to resolve registered nodes and check if all of them are unneeded. I'd say it's the matter of responsibility: scale down should not be responsible for making sure that there are no inconsistencies in the targetSize

pkg/core/scaledown/planner/planner.go
norbertcyran: >But if budget was >0 initially and was only exhausted by a sibling node in the same atomic group, it should mark the removal as IsRisky (sequential drain needed) rather than hard-failing the atomic group

I'm afraid this is not true because of that drainability rule. Can you add a test case for that?

pkg/simulator/cluster.go
```diff
@@ -220,6 +221,69 @@ func (r *RemovalSimulator) SimulateNodeRemoval(
...
+	logger := klog.FromContext(ctx)
+
+	destinationMapCopy := make(map[string]bool, len(destinationMap))
+	for k, v := range destinationMap {
+		destinationMapCopy[k] = v
+	}
+
+	var pdbsSnapshot []*policyv1.PodDisruptionBudget
+	if remainingPdbTracker != nil {
+		pdbsSnapshot = remainingPdbTracker.GetPdbs()
```
norbertcyran, line 247: as far as I see, `GetPdbs()` does not deep copy the PDBs. If I'm not mistaken, calling `RemovePods` will mutate pdbsSnapshot too. Please add a test case to confirm/rule out

pkg/simulator/cluster_test.go
```diff
@@ -237,6 +237,114 @@ func TestSimulateNodeRemoval(t *testing.T) {
...
+	pod1.Spec.NodeName = node1.Name
+
+	pod2 := BuildTestPod("pod2", 100, 100000)
+	pod2.OwnerReferences = ownerRefs
+	pod2.Spec.NodeName = node2.Name
+
+	podUnmovable := BuildTestPod("pod-unmovable", 100, 100000)
+	podUnmovable.Spec.NodeName = node2.Name
+
+	testCases := []struct {
```
norbertcyran, line 277: please add test cases with PDBs

pkg/simulator/cluster.go
```diff
@@ -220,6 +221,69 @@ func (r *RemovalSimulator) SimulateNodeRemoval(
...
+		}
+
+		if remainingPdbTracker != nil {
+			_, inParallel, _ := remainingPdbTracker.CanRemovePods(removable.PodsToReschedule)
+			if !inParallel {
+				removable.IsRisky = true
+			}
+			remainingPdbTracker.RemovePods(removable.PodsToReschedule)
+		}
+		delete(destinationMap, removable.Node.Name)
```
norbertcyran, line 280: why are nodes from the same atomic node group in `destinationMap` in the first place? In the end, we aim to reschedule all pods running in the node group, so I don't think it makes sense to shuffle the pods between the nodes in the node group that we aim to remove, right?

pkg/simulator/cluster_test.go
```diff
@@ -237,6 +237,114 @@ func TestSimulateNodeRemoval(t *testing.T) {
...
+		name                 string
+		groupNodes           []string
+		pods                 []*apiv1.Pod
+		initialDestinations  map[string]bool
+		expectUnremovable    bool
```
norbertcyran, line 282: nit: it's idiomatic in Go to use `want` over `expected`

pkg/simulator/cluster_test.go
```diff
@@ -237,6 +237,114 @@ func TestSimulateNodeRemoval(t *testing.T) {
...
+				assert.NotNil(t, unremovable)
+				assert.Nil(t, removable)
+			} else {
+				assert.Nil(t, unremovable)
+				assert.Len(t, removable, tc.expectedRemovableLen)
+			}
+			assert.Equal(t, tc.expectedFailedIndex, failedIdx)
+			for k, expectedVal := range tc.expectedDestinations {
+				assert.Equalf(t, expectedVal, destinations[k], "destinationMap mismatch for node %s", k)
+			}
```
norbertcyran, line 343: this will pass if there are extra keys in `destinations`. Please use `cmp.Diff`

pkg/simulator/cluster_test.go
```diff
@@ -237,6 +237,114 @@ func TestSimulateNodeRemoval(t *testing.T) {
...
+			snapshot := testsnapshot.NewTestSnapshotOrDie(t)
+			clustersnapshot.InitializeClusterSnapshotOrDie(t, snapshot, []*apiv1.Node{node1, node2, destNode}, tc.pods)
+
+			destinations := make(map[string]bool, len(tc.initialDestinations))
+			for k, v := range tc.initialDestinations {
+				destinations[k] = v
+			}
+
+			r := NewRemovalSimulator(registry, snapshot, testDeleteOptions(), nil, false)
+			removable, unremovable, failedIdx := r.SimulateNodesGroupRemoval(context.Background(), tc.groupNodes, destinations, time.Now(), nil)
```
norbertcyran, line 331: nit: use `t.Context()`

pkg/test/integration/inmemory/planner_test.go
```diff
@@ -0,0 +1,489 @@
...
+		assert.Equal(t, 0, finalSizeNonAtomic, "Non-atomic empty nodes should be scaled down")
+
+		finalSizeAtomic, _ := fakes.CloudProvider.GetNodeGroup("ng-atomic").TargetSize(ctx)
+		assert.Equal(t, 0, finalSizeAtomic, "Atomic empty nodes should be scaled down")
+	})
+}
+
+// TestPlanner_AtomicNodesProcessedAfterNonAtomicLimit verifies that when non-atomic simulation
+// hits the unneededNodesLimit, non-atomic loop breaks, but atomic nodes are still evaluated.
+func TestPlanner_AtomicNodesProcessedAfterNonAtomicLimit(t *testing.T) {
```
norbertcyran, line 104: is there any guarantee that atomic node groups bypassed the limit, or they just happened to be processed before the non-atomic nodes?

pkg/test/integration/inmemory/planner_test.go
```diff
@@ -0,0 +1,489 @@
...
+
+		// Run CA loop once to mark unneeded and run simulation.
+		synctestutils.MustRunOnceAfter(t, autoscaler, plannerStepDuration)
+		synctestutils.MustRunOnceAfter(t, autoscaler, plannerTestUnneededTime+time.Second)
+
+		// Run loop to actuate.
+		synctestutils.MustRunOnceAfter(t, autoscaler, plannerStepDuration)
+
+		sizeA, _ := fakes.CloudProvider.GetNodeGroup("ng-atomic").TargetSize(ctx)
+		assert.LessOrEqual(t, sizeA, 2, "Atomic node group evaluation with PDB succeeded")
```
norbertcyran, line 281: why do we allow equal?

pkg/test/integration/inmemory/planner_test.go
```diff
@@ -0,0 +1,489 @@
...
+			fakecloudprovider.WithNodes(aTemplate, 2),
+			fakecloudprovider.WithNGOptions(atomicOpts),
+		)
+
+		// Add a non-removable pod to node 0 of ng-atomic so node 0 is NOT unneeded.
+		blockingPod := test.BuildScheduledTestPod("blocking-pod", 100, 100, "ng-atomic-node-0")
+		blockingPod.Annotations = map[string]string{
+			"cluster-autoscaler.kubernetes.io/safe-to-evict": "false",
+		}
+		fakes.K8s.AddPod(blockingPod)
```
norbertcyran, line 320: how do we know that the node group has been prefiltered? If there was no prefiltering logic, wouldn't the result of this test be the same?

pkg/test/integration/inmemory/planner_test.go
```diff
@@ -0,0 +1,489 @@
...
+		// ng-nonatomic-node-0 has a pod matching shared-pdb. It can scale down once PDB is rolled back.
+		podNA := test.SetRSPodSpec(test.BuildScheduledTestPod("pod-na-0", 100, 100, "ng-nonatomic-node-0"), "rs-na-0")
+		podNA.Labels = map[string]string{"app": "shared"}
+		fakes.K8s.AddPod(podNA)
+	})
+}
+
+// runAtomicEarlyAbortRollbackTest sets up an atomic node group (3 nodes) and a non-atomic
+// node group (1 node), invokes setupCluster to configure pods/resources, and verifies that
+// simulation early-abort prevents scale down of the atomic group while allowing the non-atomic group to scale down.
```
norbertcyran, line 402: how do we know that it aborts early? I think this test can only verify if the rollback happened, early-aborting is just an internal optimization

pkg/test/integration/inmemory/planner_test.go
```diff
@@ -0,0 +1,489 @@
...
+		_, err := fakes.KubeClient.PolicyV1().PodDisruptionBudgets("default").Create(ctx, pdb, metav1.CreateOptions{})
+		assert.NoError(t, err)
+
+		// ng-atomic-node-0 is EMPTY, so EmptySorting orders it first.
+		// ng-atomic-node-1 has a pod matching shared-pdb, consuming the 1 allowed disruption.
+		podA1 := test.SetRSPodSpec(test.BuildScheduledTestPod("pod-a-1", 100, 100, "ng-atomic-node-1"), "rs-a-1")
+		podA1.Labels = map[string]string{"app": "shared"}
+		fakes.K8s.AddPod(podA1)
+
+		// ng-atomic-node-2 has a pod matching shared-pdb. Since budget is exhausted, this triggers early abort.
```
norbertcyran, line 388: this proves the limitation I mentioned in the earlier comment, I believe

Ideally, this should not block the atomic node group from scaling down. If PDB allows even a single disruption the atomic scale down should be still possible. It would be just slower, as after eviction of the pod running on node A, eviction of the pod running on node B will have to wait until the evicted pod gets recreated on another node

pkg/utils/atomic/atomic_test.go
```diff
@@ -0,0 +1,174 @@
...
+	snapshot := testsnapshot.NewTestSnapshotOrDie(t)
+	clustersnapshot.InitializeClusterSnapshotOrDie(t, snapshot, []*apiv1.Node{nodeAtomic1, nodeAtomic2, nodeAtomic3, nodeStandard}, nil)
+
+	autoscalingCtx := &ca_context.AutoscalingContext{
+		CloudProvider:   provider,
+		ClusterSnapshot: snapshot,
+	}
+
+	testCases := []struct {
+		name      []string
```
norbertcyran, line 143: unused field

pkg/core/scaledown/planner/planner.go
```diff
@@ -52,6 +52,7 @@ type eligibilityChecker interface {
 type removalSimulator interface {
 	DropOldHints()
 	SimulateNodeRemoval(ctx context.Context, node string, podDestinations map[string]bool, timestamp time.Time, remainingPdbTracker pdb.RemainingPdbTracker) (*simulator.NodeToBeRemoved, *simulator.UnremovableNode)
+	SimulateNodesGroupRemoval(ctx context.Context, nodeNames []string, destinationMap map[string]bool, timestamp time.Time, remainingPdbTracker pdb.RemainingPdbTracker) ([]simulator.NodeToBeRemoved, *simulator.UnremovableNode, int)
```
norbertcyran, line 55: nit: `SimulateNodeGroupRemoval`

norbertcyran (changes requested):

PR #65, opened 2026-08-17
Fix CSI awareness follow up - also enable CSI awareness by default.

gnufied:
This PR fixes four issues

~1. It refactors `NewNodeInfo` to also take `CSINode` as an argument. This removes duplication in bunch of places.~
2. The 2nd issue this PR fixes is, template lookup even for cloudproviders where csi node awareness is not implemented via `TemplateNodeInfoRegistry` as discussed in - https://github.com/kubernetes/autoscaler/issues/8881
3. Third thing this PR fixes is, it implements rebalancing tests for CSI when one or more pods are removed.
4. Enable CSI attach limit awareness by default.

```release-note
Add support for Cluster autoscaler scale up using TemplateNodeInfoRegistry.
This allows CA to work for cloudproviders where TemplateNodeInfo do not explicitly 
return any CSI information. 

Also enable CSINodeAwareSchedulingEnabled feature by default.
```

**Use of AI disclaimer:**

Although AI was used to create some part of the PR. I carefully incorporated AI suggested changes only after due consideration.


Choraden:
Hey @gnufied thanks for this change!

While looking at it, I came to realize that we should change our approach to changing exported code interfaces and breaking downstream compatibility. Here's my proposal on what guidelines should be established to prevent such issues: #66. This will definitely be a larger audience discussion, so let's mention it during next SIG meeting.

gnufied:
Agreed. Is there any other consumer of this project other than autoscaler project btw? I was originally thinking, I will update the signature of all the callers in autoscaler *after* I bump CA inside the autoscaler.

At this point - I guess the only real option is to introduce a new alterative to `NewNodeInfo` and deprecate old one, if we don't want to change the signature in one go.

Choraden:
> Is there any other consumer of this project other than autoscaler project btw? 

I believe there is some of them. GKE Cluster Autoscaler for sure.

> At this point - I guess the only real option is to introduce a new alterative to NewNodeInfo and deprecate old one, if we don't want to change the signature in one go.

I believe we still are in a point where we can break the NewNodeInfo signature one last time. We just need to figure out the pattern that we want to take to support backward compatibility in future changes.

gnufied:
Okay, as such first commit is not a requirement to get other bits merged. But we discussed during the PR that implemented CSI node awareness - https://github.com/kubernetes/autoscaler/pull/8721

I could remove the first commit if we can't agree on a interface for `NewNodeInfo` function.

pkg/processors/customresources/csi_processor_test.go
```diff
@@ -258,7 +258,7 @@ func TestFilterOutNodesWithUnreadyCSIResources(t *testing.T) {
 				"ng1": createCSINode("ng1_template", []string{"driver1"}),
 			},
 			registryNodeInfos: map[string]*framework.NodeInfo{
-				"ng1": framework.NewNodeInfo(buildTestNode("ng1_registry_template", true), nil, createCSINode("ng1_registry_template", []string{"driver1", "driver2"})),
```
Choraden: Nit: this is not really related to this commit, is it?

pkg/core/static_autoscaler_csi_test.go
```diff
@@ -258,19 +273,28 @@ func TestStaticAutoscalerCSI(t *testing.T) {
...
 					MaxMemoryTotal:                 100000000,
 					ScaleUpFromZero:                true,
 					CSINodeAwareSchedulingEnabled:  true,
 					PredicateParallelism:           1,
-					MaxNodeGroupBinpackingDuration: 1 * time.Hour,
 				},
 				nodeGroups:             nodeGroups,
 				nodeStateUpdateTime:    now,
 				mocks:                  mocks,
-				optionsBlockDefaulting: false,
```
Choraden, line 273: Why do we need this change?

Choraden (commented):
This looks good. I left a question and one nit.

pkg/core/static_autoscaler_csi_test.go
gnufied: This is needed because without this, autoscaler blocks eviction if a pod has `OwnerRefs` (such as a ReplicaSet), but owner can not be found.  This is same pattern DRA tests use too.

pkg/processors/customresources/csi_processor_test.go
gnufied: This appears to be a comment on a outdated diff. I think I removed this commit from the file? Can you please check again.

pkg/processors/customresources/csi_processor.go
```diff
@@ -58,8 +58,7 @@ func (p *CSICustomResourcesProcessor) FilterOutNodesWithUnreadyResources(ctx con
 			continue
 		}
 
-		// TODO: Use TemplateNodeInfoRegistry after #8882 is merged
-		templateNodeInfo, err := ng.TemplateNodeInfo(ctx)
+		templateNodeInfo, err := getNodeInfo(ctx, autoscalingCtx, ng)
```
towca: nit: Since `getNodeInfo` is no longer specific to the DRA processor after this change, it'd be good to extract it to a shared location (e.g. a new file) instead of keeping it in the DRA processor file.

towca:
Left one nit, otherwise LGTM! Feel free to unhold if you prefer not to address the nit.

/hold
/lgtm
/approve

gnufied:
@towca I don't see the nit. May be you forgot to post the review?

pkg/processors/customresources/csi_processor_test.go
Choraden: It looks like it's still here. Commit: "Enable CSI Volume limit awareness by default".

gnufied: I fixed this. I didn't realize you were talking about the individual commit. When I dropped the refactoring changes for `NewNodeInfo`, this line has to be changed. I squashed that change in the last commit of the PR and that is why you see it here. 

I have moved it to its proper commit now.

gnufied:
@towca @Choraden can you re-add lgtm. I merely moved the hunk from last commit to a different commit.

PR #66, opened 2026-08-18
Add CODE_CONVENTIONS.md to establish API stability policy

Choraden:
**To the project maintainers and all the contributors**: Please join this thread to discuss the future guidelines for development of Cluster Autoscaler Core Library.

#### What type of PR is this?
/kind cleanup

As Cluster Autoscaler core is transitioning into a standalone library, maintaining API stability and backward compatibility for exported functions is becoming crucial for downstream consumers and integrators.

Historically, function signatures (like `NewNodeInfo`) were modified directly, which is acceptable for a standalone binary but breaks library consumers. This need for formalized conventions was highlighted during the work on #65, where signature changes caused friction.

This document establishes recommended Go patterns (Functional Options, Config Structs) and a clear deprecation cycle (minimum of two minor releases) to guide future contributions.

 N/A

```release-note
NONE
```

```docs

```

### DIFF
--- a/CODE_CONVENTIONS.md
+++ b/CODE_CONVENTIONS.md
@@ -1,0 +1,71 @@
+# Code Conventions and API Policy
+
+This document outlines the coding standards and API policies for the `cluster-autoscaler` project. As Cluster Autoscaler evolves into a reusable library, maintaining API stability and backward compatibility is crucial for downstream consumers and integrators.
+
+## API Stability Policy
+
+Directly modifying function signatures of exported functions breaks backward compatibility and disrupts downstream library consumers. To prevent this, contributors should follow these guidelines when changing or extending public APIs.
+
+### 1. Prefer Extensible Signatures
+
+When introducing new exported functions, design them to be extensible from the beginning. Avoid long lists of positional arguments.
+
+#### Pattern A: Configuration Structs
+
+Group parameters into a configuration struct. Adding fields to a struct is generally non-breaking, provided zero/nil values are handled gracefully.
+
+```go
+// NodeInfo Example
+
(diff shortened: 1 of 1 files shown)


CODE_CONVENTIONS.md
```diff
@@ -0,0 +1,65 @@
...
+```go
+// NodeInfo Example
+
+type NodeInfoConfig struct {
+    Node    *apiv1.Node
+    Pods    []*PodInfo
+    CSINode *storagev1.CSINode
+}
+
+func NewNodeInfo(cfg NodeInfoConfig) *NodeInfo
```
gnufied, line 26: How about a mix of old and new which is:

```golang
func NewNodeInfo(node *apiv1.Node, extraNodeConfig NodeInfoConfig) *NodeInfo {
}
```

This kinda keeps `NewNodeInfo` extensible enough and without a `node` object, it makes no sense to call `NewNodeInfo` I guess, so that should be an explicit parameter.

Choraden: I agree, this looks good. One nit: extraNodeConfig --> cfg - it's more idiomatic.
Let me add this mix as another option to the proposal.

BigDarkClown:
Overall - looks good. We will need to probably at some point revise the public variables / functions in multiple places - some of these really should be private and not exported. We will also need a standardized deprecation duration for these - we have some K8s guidelines, but stating them explicitly would be a great benefit.

/lgtm
/approve

Adding hold in case you want to add the option @gnufied mentioned above.

/hold

Choraden:
Thanks @BigDarkClown. I added the option @gnufied mentioned. I'll need another lgtm.

/unhold

PR #68, opened 2026-08-19
Set default scheduler verbosity equal to -v

tetianakh:
#### What type of PR is this?
/kind bug

Sets default scheduler verbosity to -1, which means that scheduler verbosity will be equal to CA verbosity by default (was 0 before)

Fixes #

```release-note
Default verbosity for scheduler logs is set to the same value as -v. 
```

```docs

```

### DIFF
--- a/pkg/config/flags/flags.go
+++ b/pkg/config/flags/flags.go
@@ -242,7 +242,7 @@ func (p *AutoscalingFlags) AddFlags(fs *pflag.FlagSet) {
 			"--max-graceful-termination-sec flag should not be set when this flag is set. Not setting this flag will use unordered evictor by default."+
 			"Priority evictor reuses the concepts of drain logic in kubelet(https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2712-pod-priority-based-graceful-node-shutdown#migration-from-the-node-graceful-shutdown-feature)."+
 			"Eg. flag usage:  '10000:20,1000:100,0:60'")
-	fs.IntVar(&p.schedulerVerbosity, "scheduler-verbosity", 0, "Specifies the verbosity threshold for logs produced by the scheduler.")
+	fs.IntVar(&p.schedulerVerbosity, "scheduler-verbosity", -1, "Specifies the verbosity threshold for logs produced by the scheduler. Should not be greater than -v flag value. If it's set to -1 the scheduler log verbosity will be equal to -v flag value.")
 }
 
 // Options builds AutoscalingOptions from the flags and returns them by value
@@ -309,15 +309,17 @@ func (p *AutoscalingFlags) Options() (config.AutoscalingOptions, error) {
 		p.o.MaxStartupTime = maxHealthCheckTimeout
 	}
 
-	var verbosity int = p.schedulerVerbosity
+	var verbosity int
 	if vFlag := flag.CommandLine.Lookup("v"); vFlag != nil {
 		if v, err := strconv.Atoi(vFlag.Value.String()); err == nil {
 			verbosity = v
 		}
 	}
-
 	if verbosity < p.schedulerVerbosity {
-		klog.Warningf("--scheduler-verbosity should be at most as high as -v flag, overriding it to: %d", verbosity)
+		klog.Warningf("--scheduler-verbosity should be at most as high as --v flag, overriding it to: %d", verbosity)
+		p.schedulerVerbosity = verbosity
+	}
+	if p.schedulerVerbosity < 0 {
 		p.schedulerVerbosity = verbosity
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/config/flags/flags.go
```diff
@@ -242,7 +242,7 @@ func (p *AutoscalingFlags) AddFlags(fs *pflag.FlagSet) {
 			"--max-graceful-termination-sec flag should not be set when this flag is set. Not setting this flag will use unordered evictor by default."+
 			"Priority evictor reuses the concepts of drain logic in kubelet(https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2712-pod-priority-based-graceful-node-shutdown#migration-from-the-node-graceful-shutdown-feature)."+
 			"Eg. flag usage:  '10000:20,1000:100,0:60'")
-	fs.IntVar(&p.schedulerVerbosity, "scheduler-verbosity", 0, "Specifies the verbosity threshold for logs produced by the scheduler.")
+	fs.IntVar(&p.schedulerVerbosity, "scheduler-verbosity", 2, "Specifies the verbosity threshold for logs produced by the scheduler.")
```
towca: What logs does the vendored scheduler code produce at `2` verbosity? I think we'd want it to log close to nothing at the default value.

tetianakh: in plugins there are 2 logs at V(2), both in volumebinding plugin. There's one log at V(3) in interpodaffinity which is logged per namespace. I guess we could set the default to 3 even. AFAIK the scheduler is run in prod with --v=2, hence the default.

tetianakh: If we're OK with logging nothing, we can just leave this at 0 and not change anything.

towca: Ah I see, so basically before we added the `--scheduler-verbosity` flag the `V()` value for scheduler code would just be CA's general `--v` value, right? At `--v=4` that we use in GKE I do recall some logs from volume-related plugins, as well as the more recent ones from the taint plugin. IMO both of these are spammy and not very useful, it probably makes sense to suppress them in most cases. But I'm not sure if we should just silently change this behavior.

We could instead keep `--v` behavior backwards-compatible by defaulting unset `--scheduler-verbosity` to `--v`. This way users have `--v` working as-is without any changes, and they can opt into silencing the scheduler logs by setting `--scheduler-verbosity` lower than `--v`. WDYT?

@BigDarkClown @x13n @jackfrancis @elmiko Do you have opinions on backwards compatibility of `--v` behavior wrt. logs from scheduler plugins?

tetianakh: I can set the default to -1. And the negative values would be interpreted as "equal to --v". WDYT?

towca: SGTM, unless negative values are valid for `--v` (I don't think so, but also couldn't find a 100% confirmation quickly). In that case, there is some function you can use to check if a given flag was explicitly passed. I couldn't find an example in CA quickly, but we definitely used to use that in the past to e.g. validate that two conflicting flags are not passed together.

x13n: I think 0 is kind of backwards compatible - CA default flag values behaves the same as before. I'd honestly just leave it like that and if anyone wants more verbose scheduler logs, they can set the flag themselves. As I pointed out in [the original PR](https://github.com/kubernetes/autoscaler/pull/9977/changes#r3638628468), having `--scheduler-verbosity` higher than `-v` will emit warnings, which is not great when defaults are used.

towca: @x13n I'd expect that most users set `--v` to some >0 value for CA, it's really difficult to support it otherwise. And for all such users defaulting `--scheduler-verbosity` to 0 would silently hide all scheduler logs, users would likely only find out when debugging a problem and noticing the logs are not there.

x13n: Fair enough, maybe we should default `--scheduler-verbosity` to `-v` then. Set `--scheduler-verbosity` to -1 statically in flag definition, then during validation check if it is -1 and if so, assign current value of `-v`.

x13n: I just realized this is exactly what @tetianakh proposed above :facepalm: I guess that means I agree with the proposed approach!

tetianakh: Done!

pkg/config/flags/flags_test.go
```diff
@@ -537,31 +537,75 @@ func TestAutoscalingFlagsAllPossible(t *testing.T) {
...
+		},
+		{
+			name:                         "scheduler verbosity is unset",
+			flags:                        []string{"-v=5"},
+			wantSchedulerVerbosityOffset: 0,
+		},
+		{
+			name:                         "verbosity is unset",
+			flags:                        []string{"--scheduler-verbosity=3"},
+			wantSchedulerVerbosityOffset: 2,
```
x13n: Why is this 2 in this case? Doesn't `-v` default to 0?

tetianakh: It behaved as if the default value in tests is 5.  I thought it's WAI, but turns out that verbosity is a global variable in klog and it leaks between tests. Fixed

PR #69, opened 2026-08-19
Scaledown strategies

viktoriaas:
#### What type of PR is this?
/kind feature

This PR is a follow-up on #18 and introduces the actual Scaledown strategy framework, its primitives, engine and integration into existing CA code.

The main objective is to rank and sort scaledown candidate nodes marked by `Planner` based on configured `Strategy` (defined by dimension and weights) before they enter loop simulating scaledown candidate removal in `categorizeNodes` function. 

During a scaledown, the CA evaluates the cluster nodes to identify nodes that are underutilized and eligible for removal. Basic heuristics and defined rules are applied to filter scaledown candidates. Then, CA iterates over the list of scaledown candidates and simulates their removal in no particular order. Therefore, if a timeout happens or enough candidates have been found, iteration is stopped. It is not ensured that the optimal candidates were evaluated first and thus were at lower risk of being struck by timeout or other limits which may result in a suboptimal list of final nodes marked for scaledown. Sorted list will ensure "more suitable" candidates are evaluated first.

#### Structure
The scaledown strategy framework is introduced in its own `strategy` package under `scaledown` and includes:
##### interface.go
Defines primitives (structs and interfaces) that represent:
1. *Scoring Dimension* - a single characteristic of a Node that can be computed, normalized, and used as a contributing factor to total node score.
 2. *Node rank* - A structure that holds raw dimension scores, normalized dimension scores, weighted dimension scores and total score per each node

The design and implementation is heavily influenced by [Score phase of Kubernetes Scheduler Framework (SF)](https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/#scoring). Semantically same or similar structures are reused in CA for storing scored, normalized and weighted results for each Node as are used in the `Score` phase in K8s SF to record scoring, normalization and weighting results for each node for each scoring plugin ([NodeScoreList](https://github.com/kubernetes/kubernetes/blob/52ba90138eb40cab0987dac73e05c838149bdd1c/staging/src/k8s.io/kube-scheduler/framework/interface.go#L276), [NodePluginScores](https://github.com/kubernetes/kubernetes/blob/52ba90138eb40cab0987dac73e05c838149bdd1c/staging/src/k8s.io/kube-scheduler/framework/interface.go#L285), [PluginScore](https://github.com/kubernetes/kubernetes/blob/52ba90138eb40cab0987dac73e05c838149bdd1c/staging/src/k8s.io/kube-scheduler/framework/interface.go#L300), [PlacementPluginScore](https://github.com/kubernetes/kubernetes/blob/52ba90138eb40cab0987dac73e05c838149bdd1c/staging/src/k8s.io/kube-scheduler/framework/interface.go#L307)).  

To evaluate scaledown candidate nodes efficiently,  scoring relies on a dual-perspective data model that mimics the K8s SF. Data is grouped simultaneously by **Dimension** (column) and **Node** (row):
- `NodeDimensionScoreList `(One column in the matrix) represents one scoring Dimension, containing a list of `NodeDimensionScore` structs (the scores for every node evaluated by that dimension).
- `NodeRank` (One row in the matrix) represents one Node containing a list of raw `DimensionScore` (the evaluated scores from every dimension for that node), a list of normalized `DimensionScore` and the final aggregated `TotalScore` after weighting.

| **Node** | **Dimension** 1: Util | **Dimension** 2: PodCount |
| :--- | :--- | :--- |
| **Node** A<br>NodeRank | Column Op:<br>`NodeDimensionScore{NodeName:A, Score:80}`<br><br>Row Op:<br>`DimensionScore{DimensionName:Util, Score:80}` | Column Op:<br>`NodeDimensionScore{NodeName:A, Score:10}`<br><br>Row Op:<br>`DimensionScore{DimensionName:PodCount, Score:10}` |
| **Node** B<br>NodeRank | Column Op:<br>`NodeDimensionScore{NodeName:B, Score:20}`<br><br>Row Op:<br>`DimensionScore{DimensionName:Util, Score:20}` | Column Op:<br>`NodeDimensionScore{NodeName:B, Score:90}`<br><br>Row Op:<br>`DimensionScore{DimensionName:PodCount, Score:90}` |

**Column operations** operate on structure `NodeDimensionScore`. The data is used during the normalization phase when the normalizer iterates top-to-bottom through a single Dimension column. Normalizer needs to know which Node it is currently analyzing (hence `NodeDimensionScore` holds NodeName). It does not need the Dimension name, because the entire list belongs to one Dimension.

**Row operations** operate on structure `DimensionScore`. The data is used during weighting and aggregation when the orchestrator iterates left-to-right across a single Node. Weighting and aggregation need to know which Dimension is currently processed to apply correct weight (hence `DimensionScore` holds `DimensionName`). The process does not need the Node name, because the entire `NodeRank` struct belongs to one node.

##### ranking_session.go
Introduces engine orchestrating the Dimension scoring, normalizing, weighting and aggregating for scaledown candidates and performs final sorting. The `RankingSession` is a transient, short-lived object instantiated to hold calculations for a single evaluation cycle and statically references configured `Strategy`.

If any phase of node ranking fails (score, normalize, weight), original unsorted list is used for scaledown candidate node removal simulation.

##### strategy.go
A simple, stateless, immutable structure that holds instantiated Dimensions to evaluate and their weights. Passed in format `dimension1=weight1,dimension2=weight2,...`. Currently single scaledown strategy applies to whole cluster.

To avoid hardcoding recognized Dimensions directly into the parsing logic, the Dimensions register themselves into a package-scoped registry. A registry maps Dimension names directly to their instantiated struct instances. Individual Dimensions register themselves into the registry in `init()`.

##### utilization.go
The first implementation of the Dimension that can be expressed in a scaledown strategy as a weighted Dimension to evaluate for final score. The Dimension must comply with the core `ScoringDimension` interface to be even considered as an operational Dimension so it has to implement two functions, `Name, Score`. If the raw metric (e.g., here a utilization percentage) does not automatically align with the broader concept of "deletion preference" or fall within allowed values range, a Dimension may opt-in to the orchestrator's normalization phase by fulfilling the optional `DimensionNormalizer` interface , exposing a single `Normalize` function designed to sweep through all evaluated scores for the Dimension in that CA loop.

Objective of `UtilizationDimension` is to score the nodes and assign raw utilization as raw achieved score. It implements optional normalization to normalize raw scores to be in the allowed range and invert the meaning of raw scores which guarantees that the sorting will favor emptying and deleting the most idle nodes

#### Integration in CA

The strategy evaluation is integrated in the scaledown `Planner` as it is tightly scoped only to scaledown. 

Existing unit and benchmark tests were modified to accommodate the strategy requirement in the `Planner` structure. For the standard static autoscaler and `Planner` unit tests, an empty strategy is created and passed down simply to satisfy the updated constructor without changing the underlying test behavior. Scaledown efficiency benchmark introduced in #18 was updated to initialize and pass an actual `Strategy` configuration. The benchmark currently evaluates a hardcoded configuration with only `UtilizationDimension`, meaning the specific weight assigned to that dimension does not influence the final ranking outcomes but still sorts the list of scaledown candidate nodes. 

New flag `-scaledown-strategy` is added to pass down configured dimensions and weights to core CA engine. The flag is by default set to empty string. This ensures that default scaledown behaviour is not changed unless strategy is explicitly configured. 

```docs
- [Benchmarking scaledown strategies]: https://github.com/kubernetes-sigs/cluster-autoscaler/pull/18
```

### DIFF
--- a/pkg/core/scaledown/planner/planner_test.go
+++ b/pkg/core/scaledown/planner/planner_test.go
@@ -18,11 +18,11 @@ package planner
 
 import (
 	"fmt"
-
-	"sigs.k8s.io/cluster-autoscaler/pkg/resourcequotas"
-
 	"testing"
 	"time"
+
+	"sigs.k8s.io/cluster-autoscaler/pkg/core/scaledown/strategy"
+	"sigs.k8s.io/cluster-autoscaler/pkg/resourcequotas"
 
 	"github.com/stretchr/testify/assert"
 
@@ -509,7 +509,8 @@ func TestUpdateClusterState(t *testing.T) {
 			clustersnapshot.InitializeClusterSnapshotOrDie(t, autoscalingCtx.ClusterSnapshot, tc.nodes, tc.pods)
 			deleteOptions := options.NodeDeleteOptions{}
 			factory := resourcequotas.NewTrackerFactory(resourcequotas.TrackerOptions{CustomResourcesProcessor: processors.CustomResourcesProcessor, QuotaProvider: resourcequotas.NewCloudMinProvider(provider)})
-			p := New(&autoscalingCtx, processors, deleteOptions, nil, factory)
+			sdStrategy, _ := strategy.NewStrategy("")
+			p := New(&autoscalingCtx, processors, deleteOptions, nil, factory, sdStrategy)
 			p.eligibilityChecker = &fakeEligibilityChecker{eligible: asMap(tc.eligible)}
 			if tc.isSimulationTimeout {
 				autoscalingCtx.AutoscalingOptions.ScaleDownSimulationTimeout = 1 * time.Second
@@ -708,7 +709,8 @@ func TestUpdateClusterStatUnneededNodesLimit(t *testing.T) {
 			clustersnapshot.InitializeClusterSnapshotOrDie(t, autoscalingCtx.ClusterSnapshot, nodes, nil)
 			deleteOptions := options.NodeDeleteOptions{}
 			factory := resourcequotas.NewTrackerFactory(resourcequotas.TrackerOptions{CustomResourcesProcessor: processors.CustomResourcesProcessor, QuotaProvider: resourcequotas.NewCloudMinProvider(provider)})
-			p := New(&autoscalingCtx, processors, deleteOptions, nil, factory)
+			sdStrategy, _ := strategy.NewStrategy("")
+			p := New(&autoscalingCtx, processors, deleteOptions, nil, factory, sdStrategy)
 			p.eligibilityChecker = &fakeEligibilityChecker{eligible: asMap(nodeNames(nodes))}
 			p.minUpdateInterval = tc.updateInterval
 			p.unneededNodes.Update(&autoscalingCtx, previouslyUnneeded, time.Now())
@@ -845,7 +847,8 @@ func TestNewPlannerWithExistingDeletionCandidateNodes(t *testing.T) {
 
 			deleteOptions := options.NodeDeleteOptions{}
 			factory := resourcequotas.NewTrackerFactory(resourcequotas.TrackerOptions{CustomResourcesProcessor: processors.CustomResourcesProcessor, QuotaProvider: resourcequotas.NewCloudMinProvider(provider)})
-			p := New(&autoscalingCtx, processors, deleteOptions, nil, factory)
+			sdStrategy, _ := strategy.NewStrategy("")
+			p := New(&autoscalingCtx, processors, deleteOptions, nil, factory, sdStrategy)
 
 			p.unneededNodes.AsList()
 		})
@@ -1111,7 +1114,8 @@ func TestNodesToDelete(t *testing.T) {
 			clustersnapshot.InitializeClusterSnapshotOrDie(t, autoscalingCtx.ClusterSnapshot, allNodes, nil)
 			deleteOptions := options.NodeDeleteOptions{}
 			factory := resourcequotas.NewTrackerFactory(resourcequotas.TrackerOptions{CustomResourcesProcessor: processors.CustomResourcesProcessor, QuotaProvider: resourcequotas.NewCloudMinProvider(provider)})
-			p := New(&autoscalingCtx, processors, deleteOptions, nil, factory)
+			sdStrategy, _ := strategy.NewStrategy("")
+			p := New(&autoscalingCtx, processors, deleteOptions, nil, factory, sdStrategy)
 			p.latestUpdate = time.Now()
 
 			if tc.minLimits != nil {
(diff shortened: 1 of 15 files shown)


pkg/core/scaledown/strategy/ranking_session.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+// Sort returns a sorted copy of unneeded node names according to their TotalScore in descending order.
+func (rs *RankingSession) Sort(nodeNames []string) []string {
+	sorted := append([]string(nil), nodeNames...)
+	sort.Slice(sorted, func(i, j int) bool {
```
LucasAndFlores, line 127: I'd change this to something like `slices.SortFunc`, according to [the docs](https://pkg.go.dev/sort#Slice), it runs faster and it is more ergonomic

pkg/core/scaledown/strategy/interface.go
```diff
@@ -0,0 +1,62 @@
...
+
+package strategy
+
+import (
+	"sigs.k8s.io/cluster-autoscaler/pkg/simulator/framework"
+)
+
+// NodeDimensionScore represents a score given to the specific node NodeName by a single Dimension.
+// Used for normalization.
+type NodeDimensionScore struct {
```
LucasAndFlores, line 25: I'd suggest keep the structs and interfaces where they are used. Right now, they are not exported and we don't have another strategies implemented, so keep this file is maybe a overload of back/forth to check where is implemented/where is referenced

pkg/core/scaledown/strategy/ranking_session.go
```diff
@@ -0,0 +1,143 @@
+/*
```
LucasAndFlores, line 1: For all the new files in this directory, like this one, strategy and utilization, I'd suggest implement a few unit tests, at least covering one happy path and one error scenario. I'd implement way more, because I think it makes the code easier to understand in terms of behavior and you can catch a few edge cases. And if you are using this repo to e2e test, I also would implement one covering the new behavior

pkg/core/scaledown/planner/planner_test.go
```diff
@@ -708,7 +709,8 @@ func TestUpdateClusterStatUnneededNodesLimit(t *testing.T) {
 			clustersnapshot.InitializeClusterSnapshotOrDie(t, autoscalingCtx.ClusterSnapshot, nodes, nil)
 			deleteOptions := options.NodeDeleteOptions{}
 			factory := resourcequotas.NewTrackerFactory(resourcequotas.TrackerOptions{CustomResourcesProcessor: processors.CustomResourcesProcessor, QuotaProvider: resourcequotas.NewCloudMinProvider(provider)})
-			p := New(&autoscalingCtx, processors, deleteOptions, nil, factory)
+			sdStrategy, _ := strategy.NewStrategy("")
```
LucasAndFlores, line 712: Same here, I'd implement at least one test where you are using the new package. It will help other people understand how this is used.

LucasAndFlores (commented):
Hey @viktoriaas, I took a look into your changes and read your description, good job, really interesting and I'd love to see in action (I'll try it soon). I left a few comments with suggestions with my POV, but I'd wait for @elmiko and @BigDarkClown reviews.

PR #70, opened 2026-08-21
kwok/charts: add minimalist Helm chart setup for tests

Choraden:
#### What type of PR is this?
/kind cleanup

This change adds minimalist Helm chart setup for kwok. The files mostly come from kubernetes/autoscaler/cluster-autoscaler, however they were stripped from various templating often related to different cloudproviders. This is intended to be used in E2E core CA tests. 
I tested it manually on my local kind + kwok setup. 

N/A

```release-note
NONE
```

```docs

```

### DIFF
--- a/kwok/charts/templates/deployment.yaml
+++ b/kwok/charts/templates/deployment.yaml
@@ -1,0 +1,48 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- toYaml .Values.commonLabels | nindent 4 }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      {{- toYaml .Values.commonLabels | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- toYaml .Values.commonLabels | nindent 8 }}
+    spec:
+      priorityClassName: system-cluster-critical
+      hostNetwork: true
+      dnsPolicy: Default
+      {{- if .Values.tolerations }}
+      tolerations:
+        {{- toYaml .Values.tolerations | nindent 8 }}
+      {{- end }}
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+      serviceAccountName: {{ .Release.Name }}
+      containers:
+        - image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          name: cluster-autoscaler
+          resources:
+            requests:
+              cpu: 100m
+              memory: 300Mi
+          command:
+            - /cluster-autoscaler-kwok
+            {{- range $key, $value := .Values.extraArgs }}
+            - --{{ $key }}={{ $value }}
+            {{- end }}
+          imagePullPolicy: "{{ .Values.image.pullPolicy }}"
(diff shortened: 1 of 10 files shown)


towca:
@mtrqq Could you do the first-pass review here?

kwok/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+      serviceAccountName: {{ .Release.Name }}
+      containers:
+        - image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          name: cluster-autoscaler
+          resources:
+            requests:
+              cpu: 100m
+              memory: 300Mi
+          command:
+            - /cluster-autoscaler-kwok
```
mtrqq, line 38: How the binary is getting built for this image? Is it a standalone process from regular autoscaler image? What confused me is the name of the binary

kwok/charts/values.yaml
```diff
@@ -0,0 +1,12 @@
+image:
+  repository: cluster-autoscaler-kwok
+  tag: dev
+  pullPolicy: IfNotPresent
+
+tolerations: []
+
+extraArgs: {}
```
mtrqq, line 8: Don't we want to set some basic settings in here like cloud provider and other overrides so that we don't need to specify them for every deployment? Or we'll only have 1 centralized way to deploy it for all the E2E tests?

kwok/charts/values.yaml
```diff
@@ -0,0 +1,12 @@
+image:
+  repository: cluster-autoscaler-kwok
+  tag: dev
+  pullPolicy: IfNotPresent
```
mtrqq, line 4: Do we want to use Always here to be sure that the latest version is always pulled from the registry?

mtrqq (changes requested):
Sure, will hapily do a pass!

/assign

kwok/charts/templates/deployment.yaml
Choraden: It is built via `make build-kwok` and `make image-kwok` this will be built as a e2e job step.

kwok/charts/values.yaml
Choraden: We no longer need to specify cloud provider as kwok is the only available option :).
The point of this config is to keep it simple and add possible flags from the e2e setup. 
For example we would need to add for setup that tests the bypass scheduler feature: 
```
- --bypassed-scheduler-names=non-existing-bypassed-scheduler # Required for bypassed scheduler test case.
```
However, I don't see mutual flags that we would benefit from adding here now.

kwok/charts/values.yaml
Choraden: Not really, this is the kind limitation / feature. The E2E setup will manually load the image into the cluster without any registry. If we switch policy to Always it will fail, as the registry is not setup for the tests.

mtrqq: Ack, thanks for explaining!

kwok/charts/values.yaml
mtrqq: Fair enough, let's keep it minimal, as discussed offline kwok already has the configmap used to bootstrap its settings

mtrqq:
Thanks for the elegant change! LGTM-ing assuming that we are always deploying a single version of the helm-chart per cluster

/lgtm

PR #71, opened 2026-08-21
Revert proactive scale-up considering scheduled pods

pbetkier:
https://github.com/kubernetes/autoscaler/pull/7944 introduced a bug of continuous proactive scale-up when new pods could not get created due to ResourceQuota. This is a manual revert because of repository changes in the meantime.

/kind bug

release notes:
```release-note
Proactive scale-up again only considers unscheduled pods as a valid sample to prevent continuous proactive scale-up when new pods cannot be created e.g. due to ResourceQuota
```


damikag:
/assign damikag
triage/accepted
/lgtm

PR #73, opened 2026-08-21
Remove GCE and Azure cloud providers

Choraden:
#### What type of PR is this?
/kind cleanup

These cloud providers were retained during the migration to be used for E2E tests. Since we have committed to implementing Kwok-based tests, these providers can now be safely removed to streamline repository maintenance and avoid requiring contributors to update them.

This change also removes the router provider and sets Kwok as the default cloud provider.

N/A
I had to keep gce_localssdsize provider to avoid cyclic dependency on k/a. This is to be reworked in the future while working on the library transformation. 

```release-note
NONE
```

```docs

```


pkg/config/autoscaling_options.go
```diff
@@ -61,6 +61,7 @@ type NodeGroupAutoscalingOptions struct {
 }
 
 // GCEOptions contain autoscaling options specific to GCE cloud provider.
+// TODO(Choraden): remove it.
```
mtrqq: Can we specify when we plan to remove it? e.g. once GCE CP will be moved to google managed fork

go.mod
```diff
@@ -90,6 +49,7 @@ require (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/distribution/reference v0.6.0 // indirect
 	github.com/emicklei/go-restful/v3 v3.13.0 // indirect
+	github.com/evanphx/json-patch v5.9.11+incompatible // indirect
```
mtrqq: I'm very confused on why we have new dependencies added in the context of this PR, is it because of some other PR which resulted in this change, but didn't reconcile go.mod or because kwok became default import path?

mtrqq (changes requested):

go.mod
Choraden: Good catch. When I removed it manually and run go.mod it went gone.

pkg/config/autoscaling_options.go
Choraden: Done.

Choraden:
@mtrqq I pushed another commit that fixed default cloudprovider flag value.

PR #88, opened 2026-09-01
Add --unready-nodes-scope to scope the cluster-wide unready node check

### LINKED ISSUE #10161: Unready nodes outside any node group block autoscaling of managed node groups

/area cluster-autoscaler
/kind feature

IsClusterHealthy() in pkg/clusterstate/clusterstate.go counts unready nodes across the entire cluster, including nodes that belong to no autoscaled node group:

```go
totalUnready := len(csr.totalReadiness.Unready)
if totalUnready > csr.config.OkTotalUnreadyCount &&
    float64(totalUnready) > csr.config.MaxTotalUnreadyPercentage/100.0*float64(len(csr.nodes)) {
    return false
}
```

On clusters that mix CA-managed node groups with a fleet of externally-managed nodes joined by other means, unready nodes that CA can neither reason about nor act on will halt autoscaling for the node groups it does manage. CA stops making decisions about its own capacity because of the health of capacity that was never its concern.

The impact is larger than the pause itself. The same branch in RunOnce (pkg/core/static_autoscaler.go) calls scaleDownPlanner.CleanUpUnneededNodes() before returning, which resets unneeded-since tracking. Brief external-fleet events recurring more often than --scale-down-unneeded-time will therefore starve scale-down indefinitely, while the status ConfigMap reports the cluster healthy between episodes — so the symptom is "scale-down silently never happens" rather than an obvious outage.

Tuning the existing knobs does not solve this. Because the check is a proportion of all nodes, the threshold required to be immune to the external fleet is a function of that fleet's share of the cluster. As the external fleet grows to dominate the node count, the required value approaches the point where the check can never fire at all — so there is no setting that both ignores the external fleet and leaves the check meaningful. This is made worse by external-node failures typically being correlated rather than independent: a single connectivity or control-plane event can take the whole fleet unready simultaneously, so the threshold has to clear the entire fleet rather than some fraction of it.

--max-total-unready-percentage and --ok-total-unready-count also cannot be tuned for this in isolation, because IsNodeGroupHealthy shares both and needs them to move in the opposite direction to stay useful.

A flag that scopes the cluster-wide health check to the nodes CA manages, defaulting to today's behaviour:

```sh
--unready-nodes-scope=cluster|autoscaled     (default: cluster)
```

When set to autoscaled, IsClusterHealthy() aggregates Unready and the node total over csr.perNodeGroupReadiness rather than csr.totalReadiness.

This requires no new computation and no new data structure. updateReadinessStats already populates perNodeGroup only for nodes whose NodeGroupForNode returns non-nil, while total receives every node:

```go
if nodeGroup == nil {
    // ... warnings only
} else {
    perNodeGroup[nodeGroup.Id()] = update(perNodeGroup[nodeGroup.Id()], node, nr)
}
total = update(total, node, nr)
```

So aggregating perNodeGroupReadiness yields exactly "unready nodes in autoscaled node groups", and IsNodeGroupHealthy already consumes that same structure. The change is a few lines in IsClusterHealthy plus the flag and an option field.

Happy to send the PR if the direction is acceptable.

- Raising --max-total-unready-percentage / --ok-total-unready-count — as above, the required value tracks the external fleet's share of the cluster and leaves no stable setting, and both knobs are shared with IsNodeGroupHealthy, which needs the opposite direction.
- Excluding external nodes from CA's view — no such option exists; AllNodeLister lists with labels.Everything() and nothing in the current flag set scopes it. It would also be wrong: CA needs the full node set for scale-down rescheduling simulation.

**Additional context.**:

Related but distinct: #5959 covers node groups CA does manage being entirely NotReady. This request is about nodes outside every node group.

The IsNodeGroupHealthy path shows the scoped accounting already exists and is already trusted for per-group decisions — this request is to make the same information available to the cluster-wide gate rather than to introduce anything new.

ccding:
#### What type of PR is this?

/kind feature

`IsClusterHealthy()` counts unready nodes across every node in the cluster, including nodes that belong to no autoscaled node group:

```go
totalUnready := len(csr.totalReadiness.Unready)
if totalUnready > csr.config.OkTotalUnreadyCount &&
	float64(totalUnready) > csr.config.MaxTotalUnreadyPercentage/100.0*float64(len(csr.nodes)) {
	return false
}
```

On clusters that mix autoscaled node groups with a fleet of externally managed nodes joined by other means, unready nodes that CA can neither reason about nor act on halt autoscaling for the node groups it does manage.

The impact is larger than the pause itself. The same branch in `RunOnce` calls `scaleDownPlanner.CleanUpUnneededNodes()` before returning, which resets unneeded-since tracking. External-fleet events recurring more often than `--scale-down-unneeded-time` therefore starve scale-down indefinitely, while the status ConfigMap reports the cluster healthy between episodes — so the symptom is "scale-down silently never happens" rather than an obvious outage.

The existing knobs cannot be tuned around this:

- The check is a proportion of *all* nodes, so the threshold needed to be immune to the external fleet is a function of that fleet's share of the cluster. As the fleet grows to dominate the node count, the required value approaches the point where the check can never fire at all — there is no setting that both ignores the external fleet and leaves the check meaningful.
- External-node failures are typically correlated rather than independent. A single connectivity or control-plane event can take the whole fleet unready simultaneously, so the threshold has to clear the entire fleet, not some fraction of it.
- `--max-total-unready-percentage` and `--ok-total-unready-count` are both shared with `IsNodeGroupHealthy`, which needs them to move in the opposite direction to stay useful.

This PR adds a flag that scopes the cluster-wide check to the nodes CA manages, defaulting to today's behaviour:

```
--unready-nodes-scope=cluster|autoscaled     (default: cluster)
```

When set to `autoscaled`, `IsClusterHealthy()` aggregates `Unready` and registered-node counts over `csr.perNodeGroupReadiness`, and also includes registered nodes whose node-group lookup failed.

`updateReadinessStats` distinguishes a successful `NodeGroupForNode` lookup returning no group from a failed lookup. Nodes confirmed to be outside autoscaled node groups remain excluded. Nodes whose lookup fails have unknown ownership, so their names are tracked separately and included in the scoped check.

The failed-lookup nodes use their existing classification in `totalReadiness`, preserving startup, deletion, and suspension handling. The lookup-error set is rebuilt on each readiness refresh so recovered or removed nodes do not remain in the fallback accounting. Known node groups continue to use the same readiness data consumed by `IsNodeGroupHealthy`.

Readiness collection resolves node-group ownership and the startup timeout once per registered node before updating either summary. This keeps per-group and cluster-wide classification consistent under transient lookup failures. Each node starts with the existing 15-minute fallback timeout, so nodes without a resolved group or startup options cannot inherit the preceding node group's timeout.

**Backward compatibility.** The flag defaults to `cluster`. The zero value of the new `ClusterStateRegistryConfig.UnreadyNodesScope` field also selects cluster-wide accounting, so downstream library consumers retain that scope without changes. The correction to startup-time classification applies to both scopes.

**Scope.** `IsNodeGroupHealthy` is unchanged. The `ClusterWide` node counts in the status ConfigMap continue to report whole-cluster inventory — only the health verdict is scoped — since `api.ClusterHealthCondition.NodeCounts` is defined as an inventory of nodes in the cluster and CA never reads the ConfigMap back to make decisions. That interaction is called out in the companion FAQ PR.

**Tests.** Regression coverage includes:

- Default and explicit scopes, flag validation, and options-to-registry wiring.
- Unready managed and external nodes, including the percentage denominator and dilution by a ready external fleet.
- Complete and partial ownership lookup failures, ready failed-lookup nodes in the denominator, and startup/deletion/suspension handling.
- Readiness refresh while failures persist, lookup recovery, and removal from Kubernetes while cloud instances remain.
- Consistent per-group and cluster-wide classification under transient lookup errors, and startup grace independent of node ordering.

Fixes kubernetes/autoscaler#10161

Documentation lives in `kubernetes/autoscaler`, which still owns `cluster-autoscaler/FAQ.md` after the core was split out. Companion PR: https://github.com/kubernetes/autoscaler/pull/10226

`clusterStateRegistryConfig()` was extracted from the body of `NewStaticAutoscaler` so that the `AutoscalingOptions` → `ClusterStateRegistryConfig` mapping is unit-testable; without it, deleting the new mapping line leaves every test green while the shipped flag does nothing. It is otherwise a behaviour-preserving move of the existing struct literal.

I am happy to adjust the flag name or the value vocabulary if maintainers prefer something else — the semantics are the part I care about.

```release-note
Added `--unready-nodes-scope=cluster|autoscaled` (default: `cluster`, matching previous behaviour). When set to `autoscaled`, the cluster-wide unready node health check that gates autoscaling counts nodes belonging to autoscaled node groups and nodes whose group lookup failed, so nodes confirmed to be outside every node group no longer halt autoscaling for the node groups Cluster Autoscaler manages. Readiness accounting also uses a consistent startup timeout per node, preventing transient lookup errors or node ordering from producing inconsistent health results.
```

```docs
- [Usage]: https://github.com/kubernetes/autoscaler/pull/10226
```

### DIFF
--- a/pkg/clusterstate/clusterstate.go
+++ b/pkg/clusterstate/clusterstate.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/cluster-autoscaler/pkg/clusterstate/api"
 	"sigs.k8s.io/cluster-autoscaler/pkg/clusterstate/scaleupfailures"
 	"sigs.k8s.io/cluster-autoscaler/pkg/clusterstate/utils"
+	"sigs.k8s.io/cluster-autoscaler/pkg/config"
 	"sigs.k8s.io/cluster-autoscaler/pkg/core/scaledown"
 	"sigs.k8s.io/cluster-autoscaler/pkg/metrics"
 	"sigs.k8s.io/cluster-autoscaler/pkg/observers/nodegroupchange"
@@ -91,6 +92,10 @@ type ClusterStateRegistryConfig struct {
 	// Minimum number of nodes that must be unready for MaxTotalUnreadyPercentage to apply.
 	// This is to ensure that in very small clusters (e.g. 2 nodes) a single node's failure doesn't disable autoscaling.
 	OkTotalUnreadyCount int
+	// UnreadyNodesScope selects which nodes IsClusterHealthy counts. config.UnreadyNodesScopeAutoscaled
+	// restricts the check to nodes belonging to an autoscaled node group. Any other value, including the
+	// empty one, keeps the default behaviour of counting every node in the cluster.
+	UnreadyNodesScope string
 }
 
 // IncorrectNodeGroupSize contains information about how much the current size of the node group
@@ -494,10 +499,19 @@ func (csr *ClusterStateRegistry) IsClusterHealthy() bool {
 	csr.Lock()
 	defer csr.Unlock()
 
-	totalUnready := len(csr.totalReadiness.Unready)
+	unready, registered := len(csr.totalReadiness.Unready), len(csr.nodes)
+	if csr.config.UnreadyNodesScope == config.UnreadyNodesScopeAutoscaled {
+		// Nodes outside of any autoscaled node group are not tracked in perNodeGroupReadiness,
+		// so aggregating it yields the same counts restricted to the nodes CA manages.
+		unready, registered = 0, 0
+		for _, readiness := range csr.perNodeGroupReadiness {
+			unready += len(readiness.Unready)
+			registered += len(readiness.Registered)
+		}
+	}
 
-	if totalUnready > csr.config.OkTotalUnreadyCount &&
-		float64(totalUnready) > csr.config.MaxTotalUnreadyPercentage/100.0*float64(len(csr.nodes)) {
+	if unready > csr.config.OkTotalUnreadyCount &&
+		float64(unready) > csr.config.MaxTotalUnreadyPercentage/100.0*float64(registered) {
 		return false
 	}
 
--- a/pkg/clusterstate/clusterstate_test.go
+++ b/pkg/clusterstate/clusterstate_test.go
@@ -471,6 +471,85 @@ func TestTooManyUnready(t *testing.T) {
 	assert.True(t, clusterstate.IsNodeGroupHealthy(context.Background(), "ng1"))
 }
 
+func TestIsClusterHealthyUnreadyNodesScope(t *testing.T) {
+	now := time.Now()
+
+	testCases := []struct {
+		name               string
+		unreadyNodesScope  string
+		ngNodesReady       bool
+		externalNodes      int
+		externalNodesReady bool
+		wantHealthy        bool
+	}{
+		{
+			name:          "unset scope counts unready nodes outside node groups",
+			ngNodesReady:  true,
+			externalNodes: 3,
+			wantHealthy:   false,
+		},
+		{
+			name:              "cluster scope counts unready nodes outside node groups",
+			unreadyNodesScope: config.UnreadyNodesScopeCluster,
+			ngNodesReady:      true,
+			externalNodes:     3,
+			wantHealthy:       false,
+		},
+		{
+			name:              "autoscaled scope ignores unready nodes outside node groups",
+			unreadyNodesScope: config.UnreadyNodesScopeAutoscaled,
+			ngNodesReady:      true,
+			externalNodes:     3,
+			wantHealthy:       true,
+		},
+		{
+			name:               "cluster scope dilutes unready autoscaled nodes with ready nodes outside node groups",
+			unreadyNodesScope:  config.UnreadyNodesScopeCluster,
+			externalNodes:      30,
+			externalNodesReady: true,
+			wantHealthy:        true,
+		},
+		{
+			name:               "autoscaled scope counts unready autoscaled nodes against autoscaled nodes only",
+			unreadyNodesScope:  config.UnreadyNodesScopeAutoscaled,
+			externalNodes:      30,
+			externalNodesReady: true,
+			wantHealthy:        false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			provider := testprovider.NewTestCloudProviderBuilder().Build()
+			provider.AddNodeGroup("ng1", 1, 10, 2)
+			var nodes []*apiv1.Node
+			for i := 0; i < 2; i++ {
+				node := BuildTestNode(fmt.Sprintf("ng1-%d", i), 1000, 1000)
+				SetNodeReadyState(node, tc.ngNodesReady, now.Add(-time.Minute))
+				provider.AddNode("ng1", node)
+				nodes = append(nodes, node)
+			}
+			for i := 0; i < tc.externalNodes; i++ {
+				node := BuildTestNode(fmt.Sprintf("external-%d", i), 1000, 1000)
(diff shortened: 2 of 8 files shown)


ccding:
/assign @Choraden

@Choraden, could you help with triage and an initial review when you have a chance?

The change adds an opt-in scope for the cluster-wide unready-node check so that unready nodes outside autoscaled node groups do not halt autoscaling of managed groups. The default preserves current behavior; tests cover the scope, percentage denominator, flag validation, and options-to-registry wiring. Feedback on the flag name and scope semantics would be especially helpful.

ArangoGutierrez (commented):
Thanks for this, scoping the cluster-wide check to the node groups CA actually manages fixes a real starvation mode on mixed clusters, and keeping `cluster` as the default leaves every existing deployment where it was.

The `autoscaled` path reuses the per-node-group readiness the registry already maintains, so it adds no new bookkeeping to `updateReadinessStats`, and moving the denominator to the same aggregate keeps the percentage meaningful rather than just shrinking the numerator. `TestIsClusterHealthyUnreadyNodesScope` pins both directions, including the case where a large ready fleet outside any node group dilutes an all-unready node group into looking healthy.

pkg/clusterstate/clusterstate.go
```diff
@@ -494,10 +499,19 @@ func (csr *ClusterStateRegistry) IsClusterHealthy() bool {
 	csr.Lock()
 	defer csr.Unlock()
 
-	totalUnready := len(csr.totalReadiness.Unready)
+	unready, registered := len(csr.totalReadiness.Unready), len(csr.nodes)
+	if csr.config.UnreadyNodesScope == config.UnreadyNodesScopeAutoscaled {
+		// Nodes outside of any autoscaled node group are not tracked in perNodeGroupReadiness,
+		// so aggregating it yields the same counts restricted to the nodes CA manages.
+		unready, registered = 0, 0
+		for _, readiness := range csr.perNodeGroupReadiness {
```
ArangoGutierrez, line 514: A node whose `NodeGroupForNode` lookup returns an error is skipped by `updateReadinessStats`, so under `autoscaled` it reaches no entry in `perNodeGroupReadiness` and drops out of both the unready count and the denominator: three unready `ng1` nodes with an erroring lookup flip `IsClusterHealthy` from false to true, where the `cluster` scope still returns false. The failure direction is the unsafe one, since a transient provider error opens the gate rather than closing it, and the comment just above reads as though only nodes outside a node group go untracked. Counting lookup failures separately, or falling back to the cluster-wide totals when `perNodeGroupReadiness` is empty while `csr.nodes` is not, would keep that path conservative.

ccding: Thanks for catching this. The `autoscaled` check now includes registered nodes whose group lookup fails, using their existing readiness classification for the unready count and including them in the denominator. Nodes confirmed to be outside autoscaled groups remain excluded. Failed-lookup tracking is rebuilt each refresh so recovered or removed nodes do not leave stale counts.

Each node's ownership and startup timeout are resolved once and shared by per-group and cluster-wide readiness. The fallback timeout is reset per node, preventing inconsistent classifications and dependence on the preceding node's group. This readiness correction applies to both scopes; `cluster` remains the default.

Regression tests cover complete and partial failures, recovery and removal, readiness exceptions, and node ordering. Relevant race-enabled tests and CI tests, verification, kwok E2E, and benchmarks pass.

Could you take another look?

ccding:
@Choraden @BigDarkClown @ArangoGutierrez, could you review the updated PR when you have a chance? The lookup-failure feedback is addressed with regression coverage, and tests, verification, kwok E2E, and benchmarks are passing. Thanks!

Issue #82, opened 2026-08-28
Adapt existing GCE test cases into new KwoK based suite


Choraden:
### Context 

- There are GCE based Cluster Autoscaler E2E tests running in the old repository ([source](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/e2e), [periodic runs results](https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/ci-kubernetes-e2e-gci-gce-autoscaling)). This tests use partially deprecated scripts, their execution is "heavy" and has long feedback loop. Moreover, they are somewhat coupled with GCE cloud provider, which is a no-go after the migration to the core cluster-autoscaler repository. 
- #74 initialized machinery to work with kind+kwok (kubernetes in docker + kubernetes without kubelet). This is to overcome the problems originating from GCE based setup. The newly added smoke test utilizes [e2e-framework](https://github.com/kubernetes-sigs/e2e-framework), however this decision is not final and we might revert back to ginkgo (which is used in GCE based tests) if we find it more useful in our case.

### Goal: Adapt existing GCE test cases into new KwoK based suite

The goal is to translate each test case from [e2e/](https://github.com/kubernetes-sigs/cluster-autoscaler/blob/main/pkg/e2e/cluster_size_autoscaling.go) to the new framework [test/e2e/](https://github.com/kubernetes-sigs/cluster-autoscaler/tree/main/test/e2e). 
As the tests itself are provider agnostic this should be relatively simple. However, if the new framework turns out impractical feel free to proceed with ginkgo.

Currently the GCE tests memorize the cluster state at the beginning (e.g. there are 3 nodes and no workload) and after each test case, it expects the cluster to be brought back to that state. Then the tests are executed sequentially, which makes it spend much time waiting for the cleanup. 
With KwoK we are not constrained by the number of actual VMs hosting nodes therefore, we should run all the tests cases in parallel in the same cluster and logically separate them with namespaces and taints. That would significantly reduce the overall job duration. I recommend to start with just one case to establish a pattern there. 

**Do not** start with DRA scenarios. These I expect to be more difficult due to additional installation of DRA related drivers.

PR #89, opened 2026-09-01
e2e: fix KWOK smoke test namespace

bobbywang-google:
#### What type of PR is this?

/kind bug

Updates the `fake-pod` definition in `TestClusterAutoscaling` to use `testEnv.EnvConf().Namespace()` instead of the hardcoded `"default"` namespace, preventing test workloads from polluting the default namespace.

Relates to #82

Run locally with:
```bash
go test -v -tags e2e -count=1 -run TestClusterAutoscaling ./test/e2e

### DIFF
--- a/test/e2e/clusterautoscaling_test.go
+++ b/test/e2e/clusterautoscaling_test.go
@@ -37,7 +37,7 @@ func TestClusterAutoscaling(t *testing.T) {
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "fake-pod",
-			Namespace: "default",
+			Namespace: testEnv.EnvConf().Namespace(),
 			Labels: map[string]string{
 				"app": "fake-pod",
 			},

### CODE REVIEW

File: test/e2e/clusterautoscaling_test.go [hunk 1]
```diff
@@ -134,7 +138,7 @@ func TestClusterAutoscaling(t *testing.T) {
 				t.Fatal(err)
 			}
 			// Delete the pod
-			_ = client.Resources().Delete(ctx, pod)
+			_ = client.Resources().Delete(ctx, newPod(cfg.Namespace()))
```
Choraden: Recreating a pod like this is error prone. Something might modify the pod in between.

test/e2e/clusterautoscaling_test.go
```diff
@@ -34,38 +34,40 @@ import (
 )
 
 func TestClusterAutoscaling(t *testing.T) {
-	pod := &corev1.Pod{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "fake-pod",
-			Namespace: "default",
```
Choraden, line 40: How about we use `testEnv.EnvConfig().Namespace()` for now and later refactor while working on further improvements to the tests.

Choraden (commented):
Thank you for your first contribution and welcome to the cluster-autoscaler project!

I left a comment on how we can fix it differently for now.

Please replace `Fixes #82` to `Relates to #82` in the PR description as the former one would close the whole issue once this PR is merged.

PR #95, opened 2026-09-02
provisioningrequest: enable batch processing

jackfrancis:
#### What type of PR is this?

/kind feature

This PR introduces a `--best-effort-atomic-batch-processing` option to ProvisioningRequest so that we can reduce the number of scale up operations (AtomicIncreaseSize) when batches of ProvisioningRequest resources land simultaneously.

Fixes #

```release-note
provisioningrequest: enable batch processing
```

```docs

```

### DIFF
--- a/pkg/processors/provreq/injector.go
+++ b/pkg/processors/provreq/injector.go
@@ -18,6 +18,7 @@ package provreq
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	apiv1 "k8s.io/api/core/v1"
@@ -45,6 +46,9 @@ type ProvisioningRequestPodsInjector struct {
 	lastProvisioningRequestProcessTime time.Time
 	checkCapacityBatchProcessing       bool
 	checkCapacityProcessorInstance     string
+	bestEffortAtomicBatchProcessing    bool
+	bestEffortAtomicMaxBatchSize       int
+	maxConcurrentUpdates               int
 }
 
 // IsAvailableForProvisioning checks if the provisioning request is the correct state for processing and provisioning has not been attempted recently.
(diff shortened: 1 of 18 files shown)


pkg/provisioningrequest/besteffortatomic/provisioning_class.go
```diff
@@ -96,84 +111,150 @@ func (o *bestEffortAtomicProvClass) Provision(
...
+	daemonSets []*appsv1.DaemonSet,
+	nodeInfos map[string]*framework.NodeInfo,
+) (*status.ScaleUpStatus, errors.AutoscalerError) {
+	logger := klog.FromContext(ctx)
+	logger.Info("Processing best-effort-atomic provisioning requests as one flattened batch", "batchSize", len(prs), "podsCount", len(unschedulablePods))
+
+	return o.provisionRequests(ctx, prs, unschedulablePods, nodes, daemonSets, nodeInfos)
+}
+
+// provisionRequests runs one all-or-nothing scale-up for all supplied ProvisioningRequests and
```
mukund-wayve, line 149: > into one all-or-nothing scale-up calculation

This may be problematic, especially for constrained resources with varying availability. Let's say I have 1000 provisioning requests each requesting 1 spot T4 node. The way I understand this it, it's going to be batched to a single 0 -> 1000 node scale up. We may be able to get 600 of them, but not 1000. In the current implementation, am I right in saying that it would mark all the 1000 scale-ups as failed? But we would want 600 of them (FIFO?) to actually be marked success.

I'm curious how the default pending-pods based scale-up happens. It should be handling similar cases right?


But if this is hard to do, we could use a smaller batch size using the config option provided. The varying spot availability also makes it hard to choose a good batch size though

pkg/processors/provreq/injector.go
```diff
@@ -198,11 +302,59 @@ func (p *ProvisioningRequestPodsInjector) Process(
 	return append(unschedulablePods, podsFromProvReq...), nil
 }
 
+// getPodsFromNextBestEffortAtomicBatch injects a batch of best-effort-atomic requests, but only
+// if the next eligible ProvisioningRequest in the regular ordering belongs to that class. The
+// returned bool reports whether the batch path handled this iteration; when it is false the
+// caller should fall back to the regular single-request path.
+func (p *ProvisioningRequestPodsInjector) getPodsFromNextBestEffortAtomicBatch(ctx context.Context) ([]*apiv1.Pod, bool, error) {
```
madiyar-wayve, line 330: As I understand batch collection selects every eligible best-effort-atomic request, across hardware requirements, without compatibility partitioning.

So combining CPU, GPU and high-memory ProvisioningRequests could produce no viable option even though each request is independently satisfiable. Is it possible to combine them based on hardware requirements?

jackfrancis: @madiyar-wayve can you check the business logic in `sameSchedulingRequirements` (new func added in the 2nd commit)? Are we getting closer to addressing your (valid!) observation?

pkg/provisioningrequest/besteffortatomic/provisioning_class.go
jackfrancis: @mukund-wayve can you check that the 3rd commit begins to address some of this? There are other atomic edge cases that are currently addressed (for example the +1000 single request is rejected due to not enough capacity and fails immediately).

We don't have a standard "if you reduce your count by 400 I will accept your request" API for these types of use cases...

pkg/processors/provreq/injector.go
madiyar-wayve: @jackfrancis, yes that looks like a good improvement. Thanks!

pkg/provisioningrequest/besteffortatomic/provisioning_class.go
madiyar-wayve: @jackfrancis thanks for addressing the comment! As I understand `admitPartialBatch` only helps when some node-group resize operations succeed. For a single VMSS, a +600 resize failure admits nothing even if +300 would succeed.

Could the implementation split a failed batch—e.g. 600 → 300 → 150—and admit the successfully provisioned subset (adaptive batching)? Since each `ProvisioningRequest` remains indivisible, this preserves its atomicity.

PR #96, opened 2026-09-03
Cover errored nodes deletion with a distinct metric

mtrqq:
#### What type of PR is this?

/kind feature

All normal deletions in cluster autoscaler are going through the scale-down deletion pipeline or deleting errored nodes while regularly checking node states. This change adds a metric which gets incremented only when we are removing errored nodes. While it has a lot of intersection with failed_node_creations_total, it provides more accurate measurement of the nodes which actually got successfully scheduled for the deletion and doesn't get limited on only scaled-up node groups. 

Example blintspot present right now: nodes which entered errored state without CA applying any operation on them and getting deleted without bumping any metrics

The potential problem of the current implementation is that there's no mechanism preventing deletion calls from happening on each and every loop, one of the potential solutions to that is changing metric type to gauge, but that way we'll loose the ability of measuring the increase rate and would be locked to knowing only the current state of cluster

One of the ways to solve it is to make a TTL cache with instance id as the key, but adding it wouln't fundamentally change anything but rather wouldn't allow to accurely asses how many deletion calls were issued

Fixes #

```release-note
Add nodes_with_create_errors_deleted_total metric tracking errored nodes deletion calls issued by cluster autoscaler
```

```docs

```

### DIFF
--- a/pkg/core/static_autoscaler.go
+++ b/pkg/core/static_autoscaler.go
@@ -1143,6 +1143,7 @@ func (a *StaticAutoscaler) deleteCreatedNodesWithErrors(ctx context.Context) {
 		} else if len(nodesToDelete) > 0 {
 			deletedAny = true
 			a.clusterStateRegistry.InvalidateNodeInstancesCacheEntry(ctx, nodeGroup)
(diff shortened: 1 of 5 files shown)


norbertcyran (approved):
/lgtm

PR #102, opened 2026-09-04
Introduce timeout for SchedulablePodGroups evaluation before bin packing

damikag:
#### What type of PR is this?
/kind feature

Introduce a timeout to bound the time spent evaluating `SchedulablePodGroups` across node groups before bin packing.

To support this timeout check effectively, the loop order has been swapped so that pod equivalence groups form the outer loop and node groups form the inner loop. If the timeout is hit mid-loop, this loop structure ensures that every node group has been evaluated against the exact same initial subset of pod equivalence groups (unless the timeout is triggered while processing the very first pod equivalence group).

When there are a large number of pod equivalence groups, exhaustive predicate checking can exceed the maxInactivityTimeout. This can cause Cluster Autoscaler to fail its liveness probe and be killed by the kubelet. Additionally, in some scenarios, bin packing is prematurely cut off after evaluating only the first expansion option because the overall bin packing timeout was exceeded.

Fixes #

```release-note
Added the `--max-schedulable-pod-group-processing-duration` flag (default: 2m) to bound the time spent evaluating schedulable pod groups across node groups. This prevents Cluster Autoscaler from hanging or excessively delaying scaling decisions in clusters with a large number of pod equivalence groups.
```

```docs

```


pkg/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -40,7 +41,7 @@ import (
 	"sigs.k8s.io/cluster-autoscaler/pkg/resourcequotas"
 	"sigs.k8s.io/cluster-autoscaler/pkg/simulator"
 	"sigs.k8s.io/cluster-autoscaler/pkg/simulator/framework"
-	"sigs.k8s.io/cluster-autoscaler/pkg/utils/errors"
+	autoscaling_err "sigs.k8s.io/cluster-autoscaler/pkg/utils/errors"
```
Choraden, line 44: Let's not use snake case in import aliases. This is against go guidelines. I know CA codebase has a lot of those, but we should stop introducing new ones. Idiomatic way would be `autoscalingerr` or feel free to look for something cleaner maybe `caerr`.

pkg/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -542,6 +543,35 @@ func (o *ScaleUpOrchestrator) processCreateNodeGroupResult(
...
 }
 
+// SchedulablePodGroupsForNodeGroups returns a map of node group IDs to the pod equivalence groups
+// that can be scheduled on them. It evaluates each pod equivalence group against every node group.
+// The evaluation is time-bounded by MaxSchedulablePodGroupsProcessingTime. If the timeout is reached,
+// the function stops processing and returns the partial results accumulated up to that point.
+func (o *ScaleUpOrchestrator) SchedulablePodGroupsForNodeGroups(ctx context.Context, podEquivalenceGroups []*equivalence.PodGroup, nodeGroups []cloudprovider.NodeGroup, nodeInfos map[string]*framework.NodeInfo) map[string][]estimator.PodEquivalenceGroup {
+	schedulablePodGroups := make(map[string][]estimator.PodEquivalenceGroup)
+	logger := klog.FromContext(ctx)
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, o.autoscalingCtx.MaxSchedulablePodGroupsProcessingTime)
```
Choraden, line 553: Nit: I don't see a reason to introduce new variable `ctxWithTimeout`. I'd prefer to just shadow `ctx`.

pkg/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -542,6 +543,35 @@ func (o *ScaleUpOrchestrator) processCreateNodeGroupResult(
 	return []nodegroups.CreateNodeGroupResult{result}, nil, nil
 }
 
+// SchedulablePodGroupsForNodeGroups returns a map of node group IDs to the pod equivalence groups
+// that can be scheduled on them. It evaluates each pod equivalence group against every node group.
+// The evaluation is time-bounded by MaxSchedulablePodGroupsProcessingTime. If the timeout is reached,
+// the function stops processing and returns the partial results accumulated up to that point.
+func (o *ScaleUpOrchestrator) SchedulablePodGroupsForNodeGroups(ctx context.Context, podEquivalenceGroups []*equivalence.PodGroup, nodeGroups []cloudprovider.NodeGroup, nodeInfos map[string]*framework.NodeInfo) map[string][]estimator.PodEquivalenceGroup {
```
Choraden, line 550: Do we need to make the function public?

pkg/core/test/common.go
```diff
@@ -174,6 +174,9 @@ func NewScaleTestAutoscalingContext(
 	processorCallbacks processor_callbacks.ProcessorCallbacks, debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter,
 	templateNodeInfoRegistry ca_context.TemplateNodeInfoRegistry,
 ) (ca_context.AutoscalingContext, error) {
+	if options.MaxSchedulablePodGroupsProcessingTime == 0 {
```
Choraden, line 177: It looks, like it doesn't belong here. Can we find a better place for it?

pkg/core/scaleup/orchestrator/orchestrator_test.go
```diff
@@ -1185,6 +1187,9 @@ func runSimpleScaleUpTest(t *testing.T, testConfig *ScaleUpTestConfig) *ScaleUpT
 	if testConfig.Options != nil {
 		options = *testConfig.Options
 	}
+	if options.MaxSchedulablePodGroupsProcessingTime == 0 {
+		options.MaxSchedulablePodGroupsProcessingTime = 2 * time.Minute
+	}
```
Choraden, line 1192: Do we need this? It is already configured via defaultOptions.

Choraden (commented):
Thanks for the change Damika! I have a couple of comments/questions.

PR #103, opened 2026-09-04
Do not propagate NodeGroupForNode error in quota tracker

### LINKED ISSUE #56: Quota tracking blocks scale ups for some cloud providers when upcoming template nodes are present

/area cluster-autoscaler

**What version of the component are you using?**:

1.36

Component version:

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: v1.35.2
Kustomize Version: v5.7.1
Server Version: v1.35.6-dd.5
</pre></details>

**What environment is this in?**:

GCP (scale ups are blocked)
AWS + Azure (scale up log warnings)

**What did you expect to happen?**:

I expected to be able to perform multiple concurrent upscales 

**What happened instead?**:

I am effectively limited to a single concurrent upscale in GCP, until all template nodes have been resolved with real nodes. Subsequent scale up loops while there are template nodes present show errors like:
```
Failed to scale up: could not create quotas tracker: failed to get node group for node "template-node-for-https://www.googleapis.com/compute/v1/projects/<project>/zones/<zone>/instanceGroups/<mig-name>-6073996095659473699-upcoming-0": wrong id: expected format gce://<project-id>/<zone>/<name>, got nil
```

It does not block scale ups on all providers; on AWS, this is a soft failure that just logs warnings instead; the scale ups are not blocked, e.g.
```
Node template-node-for-<asg>-6976803830124038895-upcoming-0 has no providerId
```

**How to reproduce it (as minimally and precisely as possible)**:

This manifests when a scale up is triggered while upcoming template nodes are present. If I understand the flow correctly
1. CA decides to upscale node group A and checks the quotas to confirm it's able to proceed
2. Quota resolution succeeds (since no template nodes are present); CA increases the size of the MIG and creates new template nodes
3. CA decides to upscale node group B on its next scale up attempt and checks the quotas to confirm it's able to proceed
4. In `usageCalculator::calculateUsages`, it attempts to resolve the associated nodeGroup with `autoscalingCtx.CloudProvider.NodeGroupForNode(node)`; this implementation varies for cloud provider, and in GCP is dependent on the node having a valid provider ID, which template nodes do not have. This fails, and the error bubbles up the chain causing calculateUsages -> newQuotaTracker -> ScaleUp to fail.
```go
func (gce *GceCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.NodeGroup, error) {
    ref, err := GceRefFromProviderId(node.Spec.ProviderID)
    if err != nil {
        klog.Errorf("Error extracting node.Spec.ProviderID for node %v: %v", node.Name, err)
        return nil, err
    }
    ...
```
(this varies per provider, e.g. AWS just returns `nil, nil`)
```go
func (aws *awsCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.NodeGroup, error) {
    if len(node.Spec.ProviderID) == 0 {
        klog.Warningf("Node %v has no providerId", node.Name)
        return nil, nil
    }
    ...
```

**Anything else we need to know?**:

It seems reasonable to be including the template nodes towards quota tracking (without it, multiple upscales from templates before the templates are replaced with real nodes might exceed a quota). The manifestation is worse on GCP because it blocks scale ups, but this implies quota tracking would be incorrect on other providers since template nodes do not provider IDs

### LINKED ISSUE #10140: GCE provider: NodeGroupForNode returns an error for non-gce:// providerIDs, breaking CA in mixed-provider clusters

/area cluster-autoscaler
/area provider/gce

**What version of the component are you using?**:

Component version: 1.32, 1.34, and 1.35.0 (all reproduce, with the failure point moving between versions)

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: v1.33.0
Server Version: v1.36.2+k3s1
</pre></details>

**What environment is this in?**:

A self-hosted k3s cluster whose nodes span multiple providers:

- control plane: Raspberry Pi with `providerID: k3s://raspberrypi-0`
- workers: GCE MIG instances (`gce://...`), plus nodes from other clouds (`azure://...`, `aws:///...`)

Cluster Autoscaler runs in-cluster, one deployment per cloud provider (`--cloud-provider=gce` with `--nodes=0:2:https://www.googleapis.com/compute/v1/projects/<project>/zones/<zone>/instanceGroups/<mig>`, alongside separate CA deployments for AWS/Azure). The AWS and Azure CA deployments work fine in this exact cluster.

**What did you expect to happen?**:

The GCE provider should treat nodes with non-`gce://` providerIDs as *not managed by this provider* and skip them, the way the AWS and Azure providers do. The `CloudProvider.NodeGroupForNode` contract says:

> should return nil if the node should not be processed by cluster autoscaler

**What happened instead?**:

`GceCloudProvider.NodeGroupForNode` parses `node.Spec.ProviderID` with `GceRefFromProviderId` and returns the parse **error** for any node whose providerID is not `gce://...`. Because callers propagate that error, a single foreign node aborts whole core loops, and CA can never reach a scale-up decision. The failure point moves between versions but the root cause is the same:

v1.32 / v1.34 (node-info building):

```
Failed to get node infos for groups: wrong id: expected format gce://<project-id>/<zone>/<name>, got k3s://raspberrypi-0
```

v1.35.0 (resource-quota tracking, right before scale-up):

```
Failed to scale up: could not create quotas tracker: failed to get node group for node "raspberrypi-0": wrong id: expected format gce://<project-id>/<zone>/<name>, got k3s://raspberrypi-0
```

(`resourcequotas/usage.go` iterates **all** cluster nodes and calls `CloudProvider.NodeGroupForNode` on each, so any non-GCE node kills the tracker.)

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a cluster where at least one node has a non-`gce://` providerID (e.g. a self-hosted control plane with `k3s://...`, or any hybrid/multi-cloud setup) and a zonal MIG providing worker nodes.
2. Run cluster-autoscaler with `--cloud-provider=gce` and a static `--nodes=min:max:<MIG url>` entry.
3. Create an unschedulable pod that targets the MIG's nodes (nodeSelector/toleration advertised via the instance template's `kube-env` `AUTOSCALER_ENV_VARS`).
4. CA logs the error above and never scales the MIG up.

**Anything else we need to know?**:

The fix appears to be a one-liner in `cloudprovider/gce/gce_cloud_provider.go`, mirroring what AWS/Azure do:

```go
func (gce *GceCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.NodeGroup, error) {
	ref, err := GceRefFromProviderId(node.Spec.ProviderID)
	if err != nil {
		// Nodes from other providers (mixed-provider clusters) are not
		// managed by this provider; per the CloudProvider contract,
		// return nil instead of an error.
		klog.V(4).Infof("Node %v has non-GCE providerID %q, treating as unmanaged", node.Name, node.Spec.ProviderID)
		return nil, nil
	}
	mig, err := gce.gceManager.GetMigForInstance(ref)
	return mig, err
}
```

I have verified this patch on 1.35.0 in the cluster described above: the fatal disappears, foreign nodes are logged and skipped, and the full chain works end-to-end (pending pod → `TriggeredScaleUp` → MIG 0→1 → node joins and pod schedules → scale-down back to 0 after the pod is gone), identical to the AWS/Azure providers in the same cluster.

This looks like a concrete instance of the contract questions being discussed in #9877 (NodeGroupForNode should return nil for nodes the provider doesn't manage), and is analogous to the EKS Hybrid Nodes report on the AWS side (#8045, closed as completed).

Happy to send a PR if the maintainers agree with this direction.

norbertcyran:
Current logic causes issue in some cloud providers, in which NodeGroupForNode can return error when called with a template node. The template node should contain all the resources already present in the capacity

#### What type of PR is this?

/kind bug

Current logic causes issue in some cloud providers (e.g. GCE), in which NodeGroupForNode can return error when called with a template node. In this case, the error is propagated up to the scale up orchestrator, which causes the scale up to fail. The template node should contain all the resources already present in the capacity, so the usage calculation logic should work just fine when called with nil NodeGroup

Fixes #56 
Fixes https://github.com/kubernetes/autoscaler/issues/10140

```release-note
Fixed quota tracker blocking the scale up when NodeGroupForNode returns an error
```

```docs

```

### DIFF
--- a/pkg/resourcequotas/usage.go
+++ b/pkg/resourcequotas/usage.go
@@ -18,9 +18,9 @@ package resourcequotas
 
 import (
 	gocontext "context"
-	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/cluster-autoscaler/pkg/context"
 )
 
@@ -66,6 +66,7 @@ func newUsageCalculator(nodeFilter NodeFilter, nodeCache *nodeResourcesCache) *u
 // calculateUsages calculates resources used by nodes for every quota.
 // Returns a map with quota ID as a key and resources used in the corresponding quota as a value.
 func (u *usageCalculator) calculateUsages(ctx gocontext.Context, autoscalingCtx *context.AutoscalingContext, nodes []*corev1.Node, quotas []Quota) (map[string]resourceList, error) {
+	logger := klog.FromContext(ctx)
 	usages := make(map[string]resourceList)
 	for _, rl := range quotas {
 		usages[rl.ID()] = make(resourceList)
@@ -78,7 +79,7 @@ func (u *usageCalculator) calculateUsages(ctx gocontext.Context, autoscalingCtx 
 
 		ng, err := autoscalingCtx.CloudProvider.NodeGroupForNode(ctx, node)
 		if err != nil {
-			return nil, fmt.Errorf("failed to get node group for node %q: %w", node.Name, err)
+			logger.Error(err, "calculateUsages: failed to get node group for node, falling back to node capacity", "node", klog.KObj(node))
 		}
 		delta, err := u.nodeCache.totalNodeResources(ctx, autoscalingCtx, node, ng)
 		if err != nil {


pkg/resourcequotas/usage.go
```diff
@@ -78,7 +78,7 @@ func (u *usageCalculator) calculateUsages(ctx gocontext.Context, autoscalingCtx
 
 		ng, err := autoscalingCtx.CloudProvider.NodeGroupForNode(ctx, node)
 		if err != nil {
-			return nil, fmt.Errorf("failed to get node group for node %q: %w", node.Name, err)
+			klog.Warningf("calculateUsages: failed to get node group for node %q, falling back to node capacity: %v", node.Name, err)
```
towca: Cluster Autoscaler codebase was recently migrated to use contextual logging (see https://github.com/kubernetes-sigs/cluster-autoscaler/blob/main/pkg/proposals/structural-contextual-logging.md), so I think this should be logged via the logger created from the context that's plumbed to this method. In particular, the structured logging has slightly different API, and there are no warnings - it's either info, or error.

towca:
Changing the error handling from propagating to just logging LGTM, left a comment for the logging itself.

pkg/resourcequotas/usage.go
norbertcyran: done, changed it to error then, as with the follow-up fixes `NodeGroupForNode` should not return error in healthy circumstances (such as if node is upcoming)

PR #104, opened 2026-09-04
DRA provider: deep copy claims from the lister before snapshotting

### LINKED ISSUE #37: DRA Provider snapshot shares ResourceClaim pointers with the informer, so a snapshot mutation writes into the informer cache

magic-peach:
#### What type of PR is this?

/kind bug

The DRA provider's `Snapshot()` stored resource claims returned directly from the lister into the snapshot's claim map, without copying them. Lister objects are shared with the informer cache and are meant to be treated as read only, but the snapshot's base layer can be mutated in place before the first `Fork()`. Storing the lister's own objects means a mutation of the snapshot claim also mutates the informer cache's copy, which every other consumer of that cache sees too. This copies each claim with `DeepCopy()` before storing it, so the snapshot owns independent objects. Added `TestProviderSnapshotDoesNotMutateListerObjects` to cover the regression directly: it mutates a claim obtained from the snapshot and asserts the object still sitting in the fake lister is unchanged.

Fixes #37

Found by reading the snapshot construction path, not from an existing issue.

```release-note
Fix a bug where the DRA snapshot could mutate resource claims still shared with the informer cache.
```

```docs

```

### DIFF
--- a/pkg/simulator/dynamicresources/provider/provider.go
+++ b/pkg/simulator/dynamicresources/provider/provider.go
@@ -55,7 +55,10 @@ func (p *Provider) Snapshot() (*drasnapshot.Snapshot, error) {
 	}
 	claimMap := make(map[drasnapshot.ResourceClaimId]*resourceapi.ResourceClaim)
 	for _, claim := range claims {
-		claimMap[drasnapshot.GetClaimId(claim)] = claim
+		// The lister returns objects shared with the informer cache, which are meant to be
+		// read-only. The snapshot's base layer can be mutated in place before the first Fork(),
+		// so it needs to own its own copies rather than the informer's.
+		claimMap[drasnapshot.GetClaimId(claim)] = claim.DeepCopy()
 	}
 
 	slices, err := p.resourceSlices.ListAll()
--- a/pkg/simulator/dynamicresources/provider/provider_test.go
+++ b/pkg/simulator/dynamicresources/provider/provider_test.go
@@ -144,6 +144,33 @@ func TestProviderSnapshot(t *testing.T) {
 	}
 }
 
+// TestProviderSnapshotDoesNotMutateListerObjects verifies that mutating a Snapshot returned by Provider.Snapshot()
+// (before any Fork()) doesn't mutate the ResourceClaim objects returned by the lister, which are shared with the
+// informer cache and are meant to be read-only.
+func TestProviderSnapshotDoesNotMutateListerObjects(t *testing.T) {
+	claim := &resourceapi.ResourceClaim{ObjectMeta: metav1.ObjectMeta{Name: "claim-1", Namespace: "default", UID: "claim-1"}}
+	origClaim := claim.DeepCopy()
+	pod := test.BuildTestPod("pod1", 1, 1, test.WithResourceClaim("a", "claim-1", ""))
+
+	claimLister := &fakeLister[*resourceapi.ResourceClaim]{objects: []*resourceapi.ResourceClaim{claim}}
+	sliceLister := &fakeLister[*resourceapi.ResourceSlice]{}
+	classLister := &fakeLister[*resourceapi.DeviceClass]{}
+	provider := NewProvider(claimLister, sliceLister, classLister)
+
+	snapshot, err := provider.Snapshot()
+	if err != nil {
+		t.Fatalf("Provider.Snapshot(): unexpected error: %v", err)
+	}
+
+	if err := snapshot.ReservePodClaims(pod); err != nil {
+		t.Fatalf("Snapshot.ReservePodClaims(): unexpected error: %v", err)
+	}
+
+	if diff := cmp.Diff(origClaim, claim); diff != "" {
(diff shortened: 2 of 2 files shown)


Choraden:
Thanks @magic-peach! 

Please add release note message briefly explaining what has been fixed. 

/lgtm
/assign @towca 

That also makes me think, shouldn't we deep copy slices and classes as well?

thc1006 (commented):
This fixes #37 as well, so `Fixes #37` in the description would let the issue close with it.

I checked it through the real `RunOnce` with the benchmark harness in `pkg/core/bench`: one node with a two-device ResourceSlice, one claim already allocated to `gpu-0` on it and reserved for nobody, one pending pod referencing the claim by name. `filterOutSchedulable` schedules that pod onto the node before any `Fork`, and on main the informer's copy of the claim comes back from the loop with a reservation for the pod; on caf6de6 it comes back empty. Two things from tracing it that may help the review: the scheduler's Reserve phase deep copies any claim it has to allocate before handing it back, so only a claim that is already allocated reaches the informer's object, and `Snapshot()` runs once per loop at `static_autoscaler.go:326` and once more per node-deletion batch in the actuator, so the copy costs about half a microsecond per claim each time on my machine.

On the slices and classes question: the snapshot never writes into those objects. `AddNodeResourceSlices` stores the list it is given and `RemoveNodeResourceSlices` drops the key, and the class layer has no writer at all, so they can stay shared as they are.

I had opened #124 for the same bug from the other side, a copy in `ensureClaimWritable` when the base layer is current, and closed it in favour of this one.

magic-peach:
@thc1006 could you please check now?

thc1006:
Yes, all good from my side, and thanks for adding Fixes #37. 7616684 is the same diff as caf6de6 on top of ee7fd73, and nothing under pkg/simulator changed on main in between. The RunOnce probe from my earlier comment passes on this head and fails on main as before.

In case someone looks at the CI benchstat: the +3% on ScaleUpDRA is runner drift. The non-DRA ScaleUp moved +2% in the same run and this code does not execute there, and an interleaved local run with 8 per arm shows no difference. The only cost I can measure is +0.5% allocs/op on ScaleDownDRA, which is the copy of its 6400 claims.

skitt:

/close

PR #105, opened 2026-09-04
DRA snapshot: fix claim mutation leak when a Pod aliases a claim twice

### LINKED ISSUE #33: DRA Snapshot Revert can leak claim mutations when a Pod references the same ResourceClaim more than once

magic-peach:
#### What type of PR is this?

/kind bug

`ensureClaimWritable` trusted the caller's own pointer once a claim id was already in the current patch, but `findPodClaims` can return the same base pointer more than once when a Pod references one claim through two aliases (`spec.resourceClaims` entries with different `name` but the same `resourceClaimName`, which the Pod API allows). The second alias then mutated the base object in place instead of the copy the first alias had already made, and `Revert` cannot undo a mutation to the base layer. `ReservePodClaims`, `UnreservePodClaims`, and the shared-claim branch of `RemovePodOwnedClaims` all share this shape. Fetches the current patched value from the patch set instead of trusting the caller when the claim id is already patched.

Added `TestReservePodClaimsAliasRevert`, confirmed it fails against the unfixed code with the exact symptom described (a leaked reservation surviving Revert) and passes with the fix.

Fixes #33

```release-note
Fix a bug where reserving a Pod's resource claims could leave a leaked reservation behind after a failed scheduling attempt was reverted, when the Pod referenced the same claim through two aliases.
```

```docs

```

### DIFF
--- a/pkg/simulator/dynamicresources/snapshot/snapshot.go
+++ b/pkg/simulator/dynamicresources/snapshot/snapshot.go
@@ -355,7 +355,13 @@ func (s *Snapshot) findPodClaims(pod *apiv1.Pod, ignoreNotTracked bool) ([]*reso
 func (s *Snapshot) ensureClaimWritable(claim *resourceapi.ResourceClaim) *resourceapi.ResourceClaim {
 	claimId := GetClaimId(claim)
 	if s.resourceClaims.InCurrentPatch(claimId) {
-		return claim
+		// A Pod can reference the same claim through more than one alias, in which case
+		// the caller's claim may still be the base layer's pointer even though this claim
+		// id already has a copy in the current patch from an earlier alias. Fetch that
+		// copy instead of trusting the caller, or an in-place mutation would land on the
+		// base object and survive a Revert.
+		current, _ := s.resourceClaims.FindValue(claimId)
+		return current
 	}
 
 	return claim.DeepCopy()
--- a/pkg/simulator/dynamicresources/snapshot/snapshot_test.go
+++ b/pkg/simulator/dynamicresources/snapshot/snapshot_test.go
@@ -722,3 +722,30 @@ func TestSnapshotForkCommitRevert(t *testing.T) {
 		compareSnapshots(t, expectedState, snapshot, "After Fork, Modify, Revert, Fork, Modify")
 	})
 }
+
+func TestReservePodClaimsAliasRevert(t *testing.T) {
+	sharedClaim := &resourceapi.ResourceClaim{ObjectMeta: metav1.ObjectMeta{Name: "shared", UID: "shared", Namespace: "default"}}
+	pod := test.BuildTestPod("aliaspod", 1, 1,
+		test.WithResourceClaim("aliasA", "shared", ""),
+		test.WithResourceClaim("aliasB", "shared", ""),
+	)
+	pod.Namespace = "default"
+
+	snapshot := NewSnapshot(map[ResourceClaimId]*resourceapi.ResourceClaim{GetClaimId(sharedClaim): sharedClaim.DeepCopy()}, nil, nil, nil)
+
+	snapshot.Fork()
+	if err := snapshot.ReservePodClaims(pod); err != nil {
+		t.Fatalf("ReservePodClaims failed: %v", err)
+	}
+	snapshot.Revert()
+
+	allClaims, err := snapshot.ResourceClaims().List()
+	if err != nil {
+		t.Fatalf("ResourceClaims().List(): unexpected error: %v", err)
+	}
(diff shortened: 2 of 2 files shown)


Choraden:
Thanks @magic-peach.

Please add release note message briefly explaining what has been fixed. 

/lgtm
/assign @towca

thc1006:
Thanks for picking this up, @magic-peach. Your `TestReservePodClaimsAliasRevert` is the two-alias case from #33; I ran it against 95e21a0 and the base layer comes back clean after Revert. With the FindValue line swapped back to `return claim` the same test leaks one reservation, so the fix is doing the work. The dynamicresources packages pass under -race too.

thc1006:
Re-checked after the rebase. eb3162d is the same diff as 95e21a0 on top of ee7fd73, and the alias test still fails on main and passes here. The fix sits in ensureClaimWritable, which all three claim mutators go through, so I also pushed the two-alias pod through UnreservePodClaims (reservation and allocation both on the base layer) and through the shared-claim branch of RemovePodOwnedClaims. Both leak on main, and the unreserve one also drops the base allocation. Both come back clean here. A nested Fork, where the caller's pointer belongs to the middle layer instead of the base, behaves the same. With #104 merged in as well, the dynamicresources packages pass under -race and the RunOnce probe from that PR stays green.

magic-peach:
@towca could you please check?

skitt:

/close

PR #106, opened 2026-09-04
Fail closed when balanced scale-up cannot evaluate quota

thc1006:
#### What type of PR is this?

/kind bug

`capScaleUpsByQuota` swallows three failures: a balanced node group with no node info is skipped silently, and errors from `CheckQuota` and `ConsumeQuota` are logged and left behind. In each case `sui.NewSize` keeps whatever delta the balancer picked, so that group is resized as though no quota applied to it. The `ConsumeQuota` one has a second effect: the quota is never reserved, so the groups after it in the slice read headroom that has already been spent.

A few lines up, `applyLimits` treats the same class of failure as fatal:

```go
checkResult, err := tracker.CheckQuota(ctx, o.autoscalingCtx, nodeGroup, nodeInfo.Node(), newNodes)
if err != nil {
    return 0, errors.ToAutoscalerError(errors.InternalError, err).AddPrefix("failed to check resource quotas: ")
}
```

So the best option stops the scale-up when its quota cannot be read, and the balanced groups scale up unlimited. This PR returns an `AutoscalerError` from `capScaleUpsByQuota` instead. `balanceScaleUps` already returns `([]nodegroupset.ScaleUpInfo, errors.AutoscalerError)` and `prepareScaleUp` already turns that into a failed scale-up status, so nothing reaches the executor and no node group is resized.

This came out of reviewing kubernetes/autoscaler#9830, where @norbertcyran agreed that letting a failed `CheckDelta` read as "not quota limited" is a bug. `capScaleUpsByQuota` is @rrangith's, from the earlier fix that made balanced scale-ups respect quotas; this only changes how it behaves when a quota cannot be evaluated, and does not touch the redistribution that #9830 adds.

None filed.

No exported signature changes. `capScaleUpsByQuota` and `balanceScaleUps` are both unexported, so this does not touch `NodeGroupSetProcessor` or anything else forks build against.

How reachable each branch is today, since it changes what the tests can show:

- The `CheckQuota` branch is reachable for a node group created during the scale-up. `applyLimits` runs before the group is created, so it warms `nodeResourcesCache` under the candidate's id, and `processCreateNodeGroupResult` then gives the group a new one. For groups that kept their id through `applyLimits` or the similar-group filter the cache already holds their resources, so `CheckQuota` cannot fail a second time.
- The missing node info branch is defensive. `processCreateNodeGroupResult` sets `nodeInfos[newId]` unconditionally across that same id change, and `balanceScaleUps` drops similar node groups with no node info before they reach here, so nothing in tree gets to it. It is still an error instead of the old `continue`, which left `sui.NewSize` at the balancer's delta and scaled the group as if no quota applied.
- The `ConsumeQuota` branch cannot be reached through the current `Tracker` at all, for that same reason: the cache is warm by the time it runs, and `ErrNegativeDelta` is ruled out by the `allowedDelta > 0` guard above it. I changed it anyway rather than leave one of the three continuing, but that is why it has no test.

The tests cover both of those branches and pin the behaviour that has to survive: `TestCapScaleUpsByQuotaSharedQuota` asserts the exact resulting sizes for two groups under one shared quota, so a regression that drops the reservation shows up as both groups taking the full headroom rather than as a vague failure.

It is separate from #103. `CheckQuota` reaches `GetNodeResourceTargets`, not `NodeGroupForNode`, so the template node error being removed there is a different path and stays benign.

```release-note
Fixed cluster-autoscaler scaling up balanced node groups without a quota check when the quota could not be evaluated.
```

### DIFF
--- a/pkg/core/scaleup/orchestrator/orchestrator.go
+++ b/pkg/core/scaleup/orchestrator/orchestrator.go
@@ -721,12 +721,13 @@ func (o *ScaleUpOrchestrator) balanceScaleUps(
 	if aErr != nil {
 		return nil, aErr
 	}
-	return o.capScaleUpsByQuota(ctx, scaleUpInfos, nodeInfos, tracker), nil
+	return o.capScaleUpsByQuota(ctx, scaleUpInfos, nodeInfos, tracker)
 }
 
 // capScaleUpsByQuota caps each group's scale-up delta by its available quota and filters
 // out groups capped to zero. Uses ConsumeQuota to commit consumed quota so subsequent groups
-// sharing the same quota see the updated limits.
+// sharing the same quota see the updated limits. A quota that cannot be evaluated is an error
+// rather than an unlimited group, as it already is in applyLimits.
 // Note: unclaimed capacity from a capped group is not redistributed to other groups;
 // a group capped below its balanced delta may leave some pods unschedulable until the
 // next autoscaler cycle.
@@ -735,7 +736,7 @@ func (o *ScaleUpOrchestrator) capScaleUpsByQuota(
 	scaleUpInfos []nodegroupset.ScaleUpInfo,
 	nodeInfos map[string]*framework.NodeInfo,
 	tracker *resourcequotas.Tracker,
-) []nodegroupset.ScaleUpInfo {
+) ([]nodegroupset.ScaleUpInfo, errors.AutoscalerError) {
 	logger := klog.FromContext(ctx)
 	for i := range scaleUpInfos {
 		sui := &scaleUpInfos[i]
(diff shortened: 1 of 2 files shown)


pkg/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -745,21 +746,21 @@ func (o *ScaleUpOrchestrator) capScaleUpsByQuota(
 		}
 		nodeInfo, found := nodeInfos[sui.Group.Id()]
 		if !found {
-			continue
+			return nil, errors.NewAutoscalerErrorf(errors.CloudProviderError, "no node info for balanced node group %s", sui.Group.Id())
```
norbertcyran, line 749: just to be sure: this branch is not expected to happen in a real scenario, right? All node groups present in the clusters should be in the map, even those created during the scale up

norbertcyran (commented):
Thanks for the fix! LGTM

/lgtm

pkg/core/scaleup/orchestrator/orchestrator.go
thc1006: You're right, and my PR description got this wrong. I wrote that a node group created during the scale-up could be missing from the map, but `processCreateNodeGroupResult` sets `nodeInfos[newId]` unconditionally: first from the candidate's template, then from the created group's own once that builds. Similar node groups with no node info are already dropped by the `!found` check further up in `balanceScaleUps`. So the branch is defensive, not reachable. The description is corrected.

The same case does still reach `CheckQuota`, which is what I conflated. `applyLimits` runs before the group is created, so it warms `nodeResourcesCache` under the candidate's id, and the created group gets a new one.

I kept the error instead of the old `continue` because a `continue` leaves `sui.NewSize` at whatever the balancer picked, so the group scales as if no quota applied to it. That is the wrong way to fail if it ever does become reachable, and `applyLimits` treats the same class of failure as fatal a few lines up.

PR #108, opened 2026-09-07
Skip fixNodeGroupSize for ZeroOrMaxNodeScaling node groups

pwitkiewicz:
#### What type of PR is this?
/kind bug

fixNodeGroupSize() decreases the cloud provider target size when the registered node count remains smaller than expected target size past MaxNodeProvisionTime.

For atomic node groups (ZeroOrMaxNodeScaling = true) partial resizing is invalid. Calling DecreaseTargetSize() attempts an illegal partial size causing mentioned function to fail and crashing the autoscaler loop with "Failed to fix node group sizes".

This change skips target size adjustments in fixNodeGroupSize() for atomic node groups, relying instead on atomic cleanup mechanisms (e.g. overrideNodesToDeleteForZeroOrMax) to preserve or tear down the group as a whole.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/core/static_autoscaler.go
+++ b/pkg/core/static_autoscaler.go
@@ -990,6 +990,14 @@ func fixNodeGroupSize(ctx context.Context, autoscalingCtx *ca_context.Autoscalin
 			return false, fmt.Errorf("failed to retrieve maxNodeProvisionTime for nodeGroup %s", nodeGroup.Id())
 		}
 		if incorrectSize.FirstObserved.Add(maxNodeProvisionTime).Before(currentTime) {
+			opts, err := nodeGroup.GetOptions(ctx, autoscalingCtx.NodeGroupDefaults)
+			if err != nil && err != cloudprovider.ErrNotImplemented {
+				return false, fmt.Errorf("failed to get options for nodeGroup %s: %v", nodeGroup.Id(), err)
+			}
+			if opts != nil && opts.ZeroOrMaxNodeScaling {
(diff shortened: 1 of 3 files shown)


pkg/core/static_autoscaler.go
```diff
@@ -990,6 +990,14 @@ func fixNodeGroupSize(ctx context.Context, autoscalingCtx *ca_context.Autoscalin
 			return false, fmt.Errorf("failed to retrieve maxNodeProvisionTime for nodeGroup %s", nodeGroup.Id())
 		}
 		if incorrectSize.FirstObserved.Add(maxNodeProvisionTime).Before(currentTime) {
+			opts, err := nodeGroup.GetOptions(ctx, autoscalingCtx.NodeGroupDefaults)
```
damikag, line 993: You are not handling the case when `err == cloudprovider.ErrNotImplemented`. Is this intentional? Why?

damikag: I believe just checking for any error works here rather than checking for specifics.

damikag:
/lgtm

Thanks for the PR.

pkg/core/static_autoscaler.go
pwitkiewicz: Hey, the `GetOptions` has a doc:

```
	// GetOptions returns NodeGroupAutoscalingOptions that should be used for this particular
	// NodeGroup. Returning a nil will result in using default options.
	// Implementation optional. Callers MUST handle `cloudprovider.ErrNotImplemented`.
```

I was simply following what I was told

PR #110, opened 2026-09-08
Modernize testing tools management with go tool  

Choraden:
#### What type of PR is this?
/kind cleanup

Migrate development and testing tools (helm, kind, setup-envtest, ginkgo)
from custom Makefile 'go install' targets to an isolated tools/ sub-module
utilizing Go 1.24+'s native 'tool' directive.

Using the native toolchain ensures that dev tools are kept perfectly in
sync with the core go version.

Fixes #

```release-note
NONE
```

```docs

```


hack/verify-go-version.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Extract Go versions using awk
+ROOT_GO_VER=$(awk '/^go [0-9]/ { print $2 }' "${KUBE_ROOT}/go.mod")
+TOOLS_GO_VER=$(awk '/^go [0-9]/ { print $2 }' "${KUBE_ROOT}/tools/go.mod")
+
+if [ "${ROOT_GO_VER}" != "${TOOLS_GO_VER}" ]; then
+  echo "Error: Go version mismatch detected!"
+  echo "  Root go.mod  : ${ROOT_GO_VER}"
+  echo "  tools/go.mod : ${TOOLS_GO_VER}"
+  echo "Please align the versions by running 'make align-go-version' or manually updating 'tools/go.mod'."
```
mtrqq: Is it some change which wasn't committed?

I would really love to avoid another makefile entry for that because patching golang version is trivial and script may not always guess the intention behind the change.

tools/go.mod
```diff
@@ -0,0 +1,212 @@
...
+	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
+	sigs.k8s.io/kind v0.32.0 // indirect
+	sigs.k8s.io/kustomize/api v0.21.1 // indirect
+	sigs.k8s.io/kustomize/kyaml v0.21.1 // indirect
+	sigs.k8s.io/randfill v1.0.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v6 v6.4.2 // indirect
+	sigs.k8s.io/yaml v1.6.0 // indirect
+)
+
+replace k8s.io/api => k8s.io/api v0.37.0-rc.0
```
mtrqq, line 148: Question, do we really need all the k8s dependencies in here? Where is it coming from?

mtrqq (changes requested):
Awesome change, I absolutely love that we don't need to install setup-envtest binary explicitly

Added couple comments, would add LGTM after resolving them

hack/verify-go-version.sh
Choraden: Good catch. I had this target, but later decided that it is not really needed. I changed the message to manually updating the go version only.

tools/go.mod
Choraden: As a library, technically, we don't. The problem arise when running kwok E2E tests. When kind had different version than CA it would cause crashes in the tests. To fix it we need to keep them in line. 
Maybe `replace` is an overkill and we could just control the indirect dependencies, but ./hack/update-deps.sh is something that we know and is battle-tested, so I decided to go this way. Feel free to propose alternatives.

Choraden:
/assign @jackfrancis

Hi Jack would you have time to look on this?

PR #114, opened 2026-09-13
loopstart: remove stale return value doc from Observer interface

magic-peach:
#### What type of PR is this?

/kind documentation

The Observer interface doc comment claimed that Refresh "returns error and a bool value whether the loop should be skipped", but Refresh(context.Context) has never returned anything. Removed the stale sentence.

Fixes #

Confirmed via git log --follow -p that the comment was already incorrect in the commit that first introduced this file, so there is no prior correct version to restore. Docs only change, verified with go build, go vet and gofmt.

```release-note
NONE
```

```docs

```


Choraden:
Thanks

/lgtm
/assign BigDarkClown

skitt:

/close

PR #115, opened 2026-09-14
loop: stop logTriggerReason from eating a pending unschedulable pod signal

magic-peach:
#### What type of PR is this?

/kind bug

logTriggerReason did a non-blocking receive on the pod observer channel purely to decide which message to log. If an unschedulable pod signal was already pending when the loop triggered for an unrelated reason, such as a scale up, scale down, or provisioning request, that receive silently drained and lost the signal. The next Wait call would then miss it and wait out the full scan interval instead of reacting immediately, with a misleading log message on top. The real, separate select at the end of Wait already handles the unschedulable pod trigger correctly, so this duplicate check in logTriggerReason was only getting in its way. Removed it, logTriggerReason now just logs the message it is given.

Fixes #

Added a regression test that reproduces a pending pod signal at the same time as another trigger reason and asserts the signal is not drained. Verified locally with go build, go vet, and go test on the loop package, all pass.

```release-note
NONE
```

```docs

```


skitt:

/close

PR #116, opened 2026-09-14
Implement marshalling for ScaleUpInfo struct

tetianakh:
#### What type of PR is this?

/kind bug

 The structured logging migration changed the final scale-up plan log call from `Infof("%v")` to a key/value pair. `fmt` calls `String()` on each element of a slice, but loggers only look for `fmt.Stringer` and `logr.Marshaler` on the logged value itself, never on the elements of a plain slice. So `[]ScaleUpInfo`  was serialized by reflecting into `ScaleUpInfo.Group`, which dropped the node  group ID and truncated the output:

        "Executing final scale-up plan" scaleUpInfos=[{"Group":{},"CurrentSize":5,...

  This adds a `ScaleUpInfos` slice type implementing both `fmt.Stringer` and `logr.Marshaler`, following the same pattern as `klog.KObjSlice`, and uses it as the type of the scale-up plan so no conversion is needed at the log calls.

        text: "Final scale-up plan" scaleUpInfos="[{ng2 1->2 (max: 10)}]"
        json: "scaleUpInfos":[{"nodeGroupId":"ng2","currentSize":1,"newSize":2,"maxSize":10}]

Go assigns implicitly between `ScaleUpInfos` and `[]ScaleUpInfo`, so no other call sites change. 

Also covers the `ScaleUpToNodeGroupMinSize` log line, which  had the same problem.

Also changes the message from "Executing final scale-up plan" back to pre-migration "Fnal scale-up plan".

Fixes #

```release-note
Fixes serialization of scale-up info in the "Final scale-up plan" log message.
```

```docs

```


norbertcyran (approved):
/lgtm

PR #118, opened 2026-09-14
Don't halt cluster-wide autoscaling when one node group TargetSize() fails

### LINKED ISSUE #10132: getTargetSizes: one node group's TargetSize() error halts scale-up/down for all node groups
**Which component are you using?**: `cluster-autoscaler`

/area cluster-autoscaler
/area provider/azure

**What version of the component are you using?**: verified on `cluster-autoscaler-release-1.30` through `-1.36`; code is byte-identical across all branches. Permalinks below pin to `cluster-autoscaler-release-1.36` tip `ea47d9b`.

**What k8s version are you using?**: not version-specific.

**What environment is this in?**: hit on Azure/AKS (VMSS provider), but the failing path is provider-agnostic.

**What did you expect to happen?**: if one node group returns an error from `TargetSize()`, CA should log it, skip that node group, and continue autoscaling the healthy ones.

**What happened instead?**: one broken node group halts *all* scale-up and scale-down cluster-wide, indefinitely, until the broken node group is removed. In our production incident this held for ~17.5 hours across dozens of healthy node groups.

**How to reproduce it**:

1. Register two node groups.
2. Make node group A's `TargetSize()` return `(0, error)` on every call.
3. Create a pending pod that fits on healthy node group B.
4. Pod stays `Pending`; CA logs `Failed to update cluster state: ...` every scan interval; B never scales up.

Real-world Azure trigger: an AKS AgentPool whose backing VMSS was never created (e.g. AKS control-plane identity lacks RBAC to provision it). `ScaleSet.getVMSSFromCache()` returns `could not find vmss: <name>` on every reconcile, freezing the whole cluster.

**Root cause**:

[`getTargetSizes`](https://github.com/kubernetes/autoscaler/blob/ea47d9bb31e4e36ffe2994d0844e49a4004916e8/cluster-autoscaler/clusterstate/clusterstate.go#L480-L490) is atomic — one node-group error discards the whole map:

```go
func getTargetSizes(cp cloudprovider.CloudProvider) (map[string]int, error) {
    result := make(map[string]int)
    for _, ng := range cp.NodeGroups() {
        size, err := ng.TargetSize()
        if err != nil {
            return map[string]int{}, err
        }
        result[ng.Id()] = size
    }
    return result, nil
}
```

Its caller [`UpdateNodes`](https://github.com/kubernetes/autoscaler/blob/ea47d9bb31e4e36ffe2994d0844e49a4004916e8/cluster-autoscaler/clusterstate/clusterstate.go#L421-L427) treats this as fatal → `updateClusterState` returns → `RunOnce` bails before any scaling code runs.

**This is inconsistent with the rest of the codebase**:

- [`Recalculate()`](https://github.com/kubernetes/autoscaler/blob/ea47d9bb31e4e36ffe2994d0844e49a4004916e8/cluster-autoscaler/clusterstate/clusterstate.go#L468-L477) calls the same `getTargetSizes` and just logs a warning on error.
- [`updateAcceptableRanges`](https://github.com/kubernetes/autoscaler/blob/ea47d9bb31e4e36ffe2994d0844e49a4004916e8/cluster-autoscaler/clusterstate/clusterstate.go#L660-L672) already tolerates missing map keys (zero on miss).
- [`ScaleUpToNodeGroupMinSize`](https://github.com/kubernetes/autoscaler/blob/ea47d9bb31e4e36ffe2994d0844e49a4004916e8/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go#L240-L250) and every other direct `TargetSize()` caller (`resource/manager.go`, `scaledown/budgets/budgets.go`, `scaledown/legacy/legacy.go`) log per-pool and `continue`.

Only `UpdateNodes` treats it as fatal, and there is no flag to change this. Managed CA offerings (AKS/GKE/EKS) don't expose CA flags, so operators hitting this have no mitigation.

**Proposed fix** — `getTargetSizes` returns partial results:

```go
func getTargetSizes(cp cloudprovider.CloudProvider) (map[string]int, error) {
    result := make(map[string]int)
    var errs []error
    for _, ng := range cp.NodeGroups() {
        size, err := ng.TargetSize()
        if err != nil {
            klog.Warningf("Failed to get target size for node group %s: %v", ng.Id(), err)
            errs = append(errs, fmt.Errorf("node group %s: %w", ng.Id(), err))
            continue
        }
        result[ng.Id()] = size
    }
    return result, errors.Join(errs...)
}
```

Existing callers already tolerate missing keys, so healthy node groups keep scaling. Happy to open a PR with a `clusterstate_test.go` test if the SIG agrees.

/sig autoscaling
/priority important-longterm

yunluyl:
#### What type of PR is this?

/kind bug

`getTargetSizes` returns an empty map on the first `TargetSize()` error, and `UpdateNodes` treats that as fatal. One broken node group then blocks scale-up and scale-down for every other node group until it is removed.

This was hit in production on Azure/AKS when an AgentPool had no backing VMSS (`could not find vmss: ...` on every reconcile) and froze the whole cluster for ~17.5 hours. The same path is provider-agnostic.

This PR:
- skips node groups whose `TargetSize()` fails, returning partial results
- logs the per-group error
- lets `UpdateNodes` continue when at least one node group succeeded
- still fails `UpdateNodes` if every node group fails (nothing useful to scale)

This matches the rest of CA (`Recalculate` and other direct `TargetSize()` callers already log-and-continue).

Supersedes https://github.com/kubernetes/autoscaler/pull/10168 (that PR targeted a deleted `cluster-autoscaler-release-1.37` branch; core logic now lives in this repo per the [split](https://github.com/kubernetes-sigs/cluster-autoscaler#cluster-autoscaler)).

Fixes https://github.com/kubernetes/autoscaler/issues/10132

The snippet in #10132 still returned `errors.Join(...)`, which `UpdateNodes` would have treated as fatal and discarded the partial map. This PR also changes `UpdateNodes` to continue when the map is non-empty.

/assign @BigDarkClown @x13n

```release-note
Fix cluster-autoscaler stalling all scale-up/down when a single node group's TargetSize() returns an error. Healthy node groups now continue to autoscale.
```

#### Additional documentation e.g., KEPs (usage docs, etc.):

```docs

```

This PR was written in part with the assistance of generative AI (Cursor).

Made with [Cursor](https://cursor.com)


BigDarkClown:
That's a great improvement, thanks @yunluyl for the contribution!

/lgtm
/approve

PR #122, opened 2026-09-16
Share E2E test logic between KWOK and GCE with parameterized TestConfig

bobbywang-google:
#### What type of PR is this?

/kind refactor

Addresses follow-up feedback from #94 by extracting common Kubernetes E2E test workflows into a shared package (`pkg/e2e/common`) and parameterizing provider-specific configuration (`TestConfig`), allowing both KWOK (`test/e2e/`) and GCE (`pkg/e2e/`) E2E test suites to execute the exact same core business logic without code duplication.

Specifically, this PR introduces:
1. **Shared Provider-Agnostic Test Workflows** (`pkg/e2e/common/workflows.go`):
   * `RunScaleDownReschedulingPodAllowedByPDB`: Verifies that CA drains a pod allowed by a PodDisruptionBudget (`minAvailable: 1`) onto an existing node and scales down the unneeded node.
   * `RunScaleDownUnneededNode`: Verifies that an unneeded node is cleanly scaled down after its workload pod is deleted.
   * `RunScaleUp`: Verifies that CA provisions a new node when a pod is unschedulable due to resource constraints.
2. **Parameterized Provider Configuration (`TestConfig`)** (`pkg/e2e/common/config.go`, `resource.go`):
   * Abstracts provider-specific node group labels (`kwok-nodegroup` vs. GCE node pool labels), tolerations (`kwok-provider:NoSchedule` vs. none), and timeouts (`NodeReadyTimeout`, `ScaleDownTimeout`, `PodSchedulingTimeout`).
   * Automatically resolves single-node allocatable CPU from live cluster nodes (`ResolveNodeCPU`) and calculates proportional CPU requests (`70%` for filler pods, `10%` and `35%` for PDB test pods) so workflows scale deterministically across both `0`-node KWOK clusters and `N`-node GCE clusters without manual node cordoning hacks.

Follow-up to #94
Relates to #82


pkg/e2e/common/resource.go
```diff
@@ -0,0 +1,204 @@
...
+
+const (
+	pauseImage = "registry.k8s.io/pause:3.9"
+)
+
+// NewTestPod creates a test pod configuration using the active TestConfig.
+// Defaults to modest baseline requests (500m CPU, 500Mi memory) when exact node-proportional
+// sizing is not required, ensuring the pod is non-BestEffort and triggers standard scheduling.
+func NewTestPod(name, namespace string) *corev1.Pod {
+	return NewTestPodWithResources(name, namespace, "500m", "500Mi")
```
Choraden: Let's parameterize available resources so that we can allocate pods with some fraction of it.

pkg/e2e/common/resource.go
```diff
@@ -0,0 +1,204 @@
...
+					Image: pauseImage,
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse(cpu),
+							corev1.ResourceMemory: resource.MustParse(memory),
+						},
+					},
+				},
+			},
+			Tolerations: cfg.Tolerations,
```
Choraden: This is good. Tolerations depend on the nodepool from given provider.

pkg/e2e/common/resource.go
```diff
@@ -0,0 +1,204 @@
...
+		},
+	}
+	if cfg.NodeGroup != "" && cfg.NodeGroupLabelKey != "" {
+		pod.Spec.NodeSelector = map[string]string{
+			cfg.NodeGroupLabelKey: cfg.NodeGroup,
+		}
+	}
```
Choraden: This is also ok. Comment as above.

pkg/e2e/common/resource.go
```diff
@@ -0,0 +1,204 @@
...
+// and performs forced node cleanup if CA didn't scale down in time (on KWOK).
+func TeardownPodAndNodeGroup(ctx context.Context, client klient.Client, pods []*corev1.Pod, nodeGroup string) {
+	cfg := GetTestConfig()
+	for _, pod := range pods {
+		if pod != nil {
+			_ = client.Resources().Delete(ctx, pod)
+		}
+	}
+	_ = WaitForPodsDeleted(ctx, client, pods, cfg.PodDeletionTimeout)
+	if cfg.Provider == ProviderKWOK {
```
Choraden: Becareful with no KwoK providers. We would like to leverage the timeouts from test configs. Adjust the values accordingly.

pkg/e2e/common/workflows.go
```diff
@@ -0,0 +1,265 @@
...
+				return true, nil
+			}
+		}
+		return false, nil
+	}, wait.WithTimeout(cfg.PodSchedulingTimeout), wait.WithContext(ctx))
+	if err != nil {
+		return fmt.Errorf("TriggeredScaleUp event not found: %v", err)
+	}
+
+	if err := WaitForPodScheduled(ctx, client, triggerPod, cfg.PodSchedulingTimeout); err != nil {
```
Choraden: Remember about adjusting PodSchedulingTimeout for non-KwoK cloud providers.

PR #123, opened 2026-09-17
CapacityBuffer: enforce ResourceQuota pod count limit

inksamurai-wq:
#### What type of PR is this?

/kind bug

Uncomments `requests[corev1.ResourcePods]` in `calculatePodRequests` (`pkg/capacitybuffer/controller/resourcequotas.go`) so that `CapacityBuffer` enforces `pods` limits in `ResourceQuota`.

Previously, `requests[corev1.ResourcePods]` was commented out, so `CapacityBuffer` completely ignored `pods` limits in `ResourceQuota`. This allowed buffers to bypass pod quotas and perform scale up.

N/A

Tested with unit tests (`TestResourceQuotaAllocator`) and manual GKE cluster verification.

```release-note
NONE
```

```docs

```


pkg/capacitybuffer/controller/resourcequotas_test.go
```diff
@@ -460,6 +460,49 @@ func TestResourceQuotaAllocator(t *testing.T) {
...
+			buffers: []*v1.CapacityBuffer{
+				testutil.NewBuffer(
+					testutil.WithStatusPodTemplateRef("empty-pod"),
+					testutil.WithStatusPodTemplateGeneration(0),
+					testutil.WithStatusReplicas(10),
+				),
+			},
+			wantReplicas:       []int32{3}, // 4 total - 1 used = 3 available
+			wantExceededQuotas: [][]string{{"quota-pods-best-effort"}},
+		},
```
walidghallab, line 513: Maybe I am missing something, but how is it different from the previous test

walidghallab:
Only one minor commit, other than that LGTM. Thank you!

Adding Norbert

/cc norbertcyran

walidghallab:
Please restore the formatting of the pull request. Pull requests in Kubernetes follow a specific format that makes it consistent with other pull requests.

Also mention why we need to enforce limit on the number of pods, which is to fix the vulnerability we discussed

pkg/capacitybuffer/controller/resourcequotas_test.go
inksamurai-wq: Yeah, having buffer with the cpu/memory definition and not checked against resource quota is same as without the definition. Thanks for the notice, I have fixed the 1 tests case, I think you'll understand what I had intended now :D

inksamurai-wq:
> Please restore the formatting of the pull request. Pull requests in Kubernetes follow a specific format that makes it consistent with other pull requests.
> 
> Also mention why we need to enforce limit on the number of pods, which is to fix the vulnerability we discussed

Done.

pkg/capacitybuffer/controller/resourcequotas_test.go
```diff
@@ -460,6 +460,57 @@ func TestResourceQuotaAllocator(t *testing.T) {
...
+					testutil.WithStatusPodTemplateRef("podTemp"),
+					testutil.WithStatusPodTemplateGeneration(0),
+					testutil.WithStatusReplicas(5),
+				),
+			},
+			wantReplicas:       []int32{3}, // CPU/memory allow 8 replicas, but pods limits to 3 (10 total - 7 used)
+			wantExceededQuotas: [][]string{{"quota-pods"}},
+		},
+		{
+			name: "buffer with zero resource requests limited by pods quota",
```
norbertcyran, line 491: nit: This test case is redundant, I think. It doesn't add any coverage on top of the above test case. Both check exactly the same thing - buffer replicas are limited by ResourceQuotas because of pod limits. Other resources do not seem to matter here. I'd suggest removing it

norbertcyran (commented):
thanks! I have one suggestion, other than that LGTM

pkg/capacitybuffer/controller/resourcequotas_test.go
inksamurai-wq: Thanks, makes sense to me, done.

PR #124, opened 2026-09-17
DRA snapshot: copy a base layer claim before writing to it

### LINKED ISSUE #37: DRA Provider snapshot shares ResourceClaim pointers with the informer, so a snapshot mutation writes into the informer cache

thc1006:
#### What type of PR is this?

/kind bug

A snapshot that was never forked has the base layer as its current patch, so `ensureClaimWritable` handed the base object itself out for in-place modification. The base holds the objects the snapshot was built from, and `Provider.Snapshot()` builds it from the lister's objects without copying (`provider.go:58`), so those are the informer's.

`RunOnce` writes at that depth twice: `filterOutSchedulable` (`static_autoscaler.go:541`) schedules the pending pods that fit onto existing nodes, and `addLatestScaleUpResultsToClusterSnapshot` (`:757`) schedules the rest onto the upcoming nodes. Neither forks, and every `Fork` elsewhere is paired with a `Revert` or a `Commit`, so both run with the base layer current. Both go through `modifyResourceClaimsForScheduledPod`. Reserve deep copies a claim it has to allocate, so those are shielded; a claim that is already allocated, which is what a shared claim referenced by name is, or one allocated before its pod went back to pending, goes straight to `ReservePodClaims` and the reservation lands on the informer's object. It exists only in the simulation, but every later snapshot starts from it until the informer replaces the object. `AddPodReservationInPlace` checks `IsReservedForPod` first, so it is one entry per pod, not growth on every loop.

The change hands a claim out for writing only when a forked layer already tracks it, and copies it otherwise, as a claim tracked only in a lower layer already was; `SetCurrent` then stores the copy in the base map, which is the snapshot's own. `PatchSet` gains `IsForked` to tell the two apart. A `DeepCopy` of a claim with one allocated device and one reservation is about 0.5 microseconds, paid once per claim written without a `Fork`.

`TestReservePodClaimsUnforked` reserves on a snapshot built from the object itself, through one reference and through two aliases; `TestSnapshotWriteStaysOutOfTheNextSnapshot` goes through `Provider.Snapshot()` twice around a write, the shape `RunOnce` has, and fails before the change because the second snapshot starts with the first one's reservation; `TestPatchSetIsForked` covers the new method across `Fork`, `Revert` and `Commit`.

Fixes #37

I reproduced it through `RunOnce` with the benchmark harness in `pkg/core/bench`: one node with a two-device ResourceSlice, one claim already allocated to `gpu-0` on it and reserved for nobody, one pending pod referencing the claim by name. On main the informer's copy of the claim comes back from the loop with a reservation for that pod; on this branch it comes back empty. The probe is not part of the PR.

#105 by @magic-peach changes the other half of this condition, the branch taken when a forked layer already tracks the claim, and the two are independent: with #105 merged in locally the resolution is that one line, this condition over her body, and her `TestReservePodClaimsAliasRevert` passes next to the tests here. `IsForked` is the same method #55 adds, with the same body, so whichever of the two lands second drops one copy.

This PR was written in part with the assistance of generative AI. I have reviewed and tested every change myself.

```release-note
Fixed the DRA snapshot writing reservations made during scale-up simulation into the informer's cached ResourceClaim objects, where they survived into later autoscaler loops.
```

```docs

```


thc1006:
Superseded by #104, which fixes the same thing at the provider and already has an lgtm.

/close

PR #127, opened 2026-09-17
Make buffer controller compatible with ctrl.Manager

norbertcyran:
CapacityBuffer controller now implements Runnable interface and handles upstream context.

#### What type of PR is this?

/kind cleanup

Makes bufferController compatible with controller-runtime Manager. This improves how bufferController handles graceful shutdown and prepares for an eventual migration to controller-runtime controller.

Fixes #

```release-note
NONE
```

```docs

```


pkg/capacitybuffer/controller/controller.go
```diff
@@ -363,7 +369,6 @@ func (c *bufferController) processNextItem() bool {
 	} else {
 		// Put the item back on the queue to handle it later
 		c.queue.AddRateLimited(key)
-		runtime.HandleError(fmt.Errorf("capacity buffer controller: error syncing namespace %q, requeueing", key))
```
walidghallab, line 366: Why removing this line?

pkg/builder/autoscaler.go
```diff
@@ -205,7 +205,10 @@ func (b *AutoscalerBuilder) Build(ctx context.Context) (core.Autoscaler, *loop.L
 			} else {
 				fakePodsResolver = fakepods.NewDefaultingResolver()
 			}
-			cbctrl.InitializeAndRunDefaultBufferController(ctx, capacitybufferClient, fakePodsResolver)
+			err = cbctrl.InitializeAndRunDefaultBufferController(ctx, b.manager, capacitybufferClient, fakePodsResolver)
```
walidghallab, line 208: Just double checking, are you sure that the ctx here will live as long as CA lives, and will be cancelled when TERM signal is passed (like in stopCh)?

walidghallab (approved):
Couple of small comments, other than that LGTM

pkg/capacitybuffer/controller/controller.go
norbertcyran: good catch, I must have removed it by accident, fixed

pkg/builder/autoscaler.go
norbertcyran: this is actually no op, as previously we passed `ctx.Done()` as `stopCh`. Though you made a very good point - it looks like we are handling this correctly in GKE CA: https://github.com/GoogleCloudPlatform/cluster-autoscaler/blob/main/main.go#L83, but we don't handle it correctly in the main repository: https://github.com/kubernetes/autoscaler/blob/b9dc0be2dd288c26df66a6b47005342314c81d9a/cluster-autoscaler/main.go#L124. I'll fix that in the main repo.

Resolving this comment since the behavior remains the same, and that's a bug in the `main.go` of the autoscaler repository

mtrqq:
Love the interface change, looks way more concise and standardized now!

/lgtm

PR #128, opened 2026-09-19
expander: add least-nodes to the list of available expanders

magic-peach:
/kind bug

AvailableExpanders in pkg/expander/expander.go, used to build the --expander flag's help text, was missing least-nodes even though it's fully registered in RegisterDefaultExpanders and has its own working implementation. The flag already worked with --expander=least-nodes, it just never showed up in --help. Added a regression test that checks AvailableExpanders against every known expander name constant so this can't silently happen again.

Fixes #

Pure documentation/help-text fix, no behavior change.

```release-note
`--expander=least-nodes` now shows up in the `--expander` flag's help text.
```

```docs

```


skitt:

/close

